PR #3, opened 2026-02-22
Load session chat history on direct navigation

MrOrz:
Implemented logic to load chat history when navigating directly to a session URL or refreshing the page. This ensures users can see past messages even if they didn't just create the session. Refactored event processing logic to be reusable for both streaming and history loading. Verified that Landing Page navigation still uses the cached data without reloading.

---
*PR created automatically by Jules for task [8634391679780268078](https://jules.google.com/task/8634391679780268078) started by @MrOrz*


MrOrz:
This PR actually works -- most changes are just prettier fix though.

PR #49, opened 2026-05-07
feat(sidebar): sort by recency, unread indicator, last-active date

### LINKED ISSUE #48: feat(sidebar): sort by recency, new-message indicator, replace ID with last-active date
## 背景

Sidebar 目前的三個問題：
1. 沒有依照時間排序，找不到最近的查核
2. 沒有提示使用者哪些 session 有新的訊息
3. 第二行顯示 session ID 無意義

### 1. 依 lastUpdateTime 排序（最新優先）
ADK 的 session list API 已回傳 `lastUpdateTime`，只需要在 client 端排序即可。

### 2. 未讀訊息 indicator
ADK 沒有原生的「已讀」機制，需要在 session `state` 存入自訂 metadata：
- 新增 `lastOpenedAt`（timestamp）存在 `session.state`，格式與現有的 `title` key 相同
- 使用者進入 session 時，PATCH session 更新 `lastOpenedAt`
- Sidebar 比對 `lastUpdateTime > lastOpenedAt`，如為 true 則顯示小圓點 indicator

### 3. 第二行改顯示最後活動日期
- 將目前的 `{session.id.slice(0, 8)}…` 替換為 `lastUpdateTime` 的格式化日期
- 未來可進一步擴充為「是否已回應」狀態或即時 progress

## 技術設計

| 功能 | ADK 原生支援 | 方案 |
|---|---|---|
| 依 lastUpdateTime 排序 | ✅ | Client-side sort |
| lastOpenedAt 追蹤 | ❌ | 存在 session.state（同 title key 的模式） |
| 最新日期顯示 | ✅ | 使用 lastUpdateTime |

## 修改範圍

- `src/lib/sessions.functions.ts` — 新增 `lastOpenedAt` key、`markSessionOpened()`
- `src/routes/_app/session.$sessionId.tsx` — session 進入時呼叫 `markSessionOpened`
- `src/components/Sidebar.tsx` — 排序、indicator dot、日期顯示取代 ID

MrOrz:
## Summary

- Sessions sorted by \`lastEventTime\` (set by agent callback) — agent activity always surfaces to the top; opening a session does **not** cause reordering
- Orange dot indicator appears when the agent has replied since the user last opened a session; dot disappears on re-entry or when new messages arrive while viewing
- Second line replaced: 8-char session ID → formatted last-active date
- New sessions initialize \`lastEventTime\` and \`lastOpenedAt\` on creation, so they sort to the top immediately without showing an unread dot
- Sessions without \`lastEventTime\` (created before this feature) sink to the bottom and never show as unread

## Technical design

**Why not \`lastUpdateTime\`?** ADK bumps \`lastUpdateTime\` on every session state PATCH, including when the client writes \`lastOpenedAt\`. Using it for sorting caused sessions to jump to the top whenever they were opened.

**\`lastEventTime\`** is written by an \`after_agent_callback\` on \`ai_writer\` (\`time.time()\`), which fires once per agent turn. Only real agent activity moves a session up. Sort logic lives in \`listSessions\` (server function) so callers get a pre-sorted list. New sessions also initialize this field on creation so they appear at the top right away.

**\`lastOpenedAt\`** is stored in \`session.state\` via \`markSessionOpened()\` PATCH (in seconds, to match Python's \`time.time()\`). Server-side storage enables cross-device unread tracking. Since sorting uses \`lastEventTime\`, this PATCH no longer affects sort order. \`markSessionOpened\` re-runs whenever \`messages.length\` changes, so messages received while the user is viewing the session are also marked as read.

**Cache freshness**: \`['sessions']\` is invalidated immediately after the SSE stream ends, so \`lastEventTime\` appears in the sidebar right after the agent responds rather than waiting for the 30-second stale window.

## Test plan

- [x] Sessions appear sorted newest-first after agent responds
- [x] Opening a session does **not** move it to the top
- [x] Navigate away, let agent respond, return to sidebar → orange dot appears
- [x] Open that session → dot disappears
- [x] Stay on session page while agent responds → no orange dot when navigating away
- [x] Create a new session → it appears at the top with no unread dot
- [x] Second line shows a human-readable date, not a raw ID
- [x] Old sessions (no \`lastEventTime\`) appear at the bottom with no dot
- [x] Edit-title mode also shows the date (not ID)

https://github.com/user-attachments/assets/a4d12f7d-330d-487b-a24e-201d1268448c

Closes #48


src/lib/sessions.functions.ts
```diff
@@ -113,3 +121,25 @@ export const updateSession = createServerFn({ method: 'POST' })
...
+      {
+        params: {
+          path: {
+            app_name: ADK_APP_NAME,
+            user_id: ADK_USER_ID,
+            session_id: sessionId,
+          },
+        },
+        body: {
+          stateDelta: { [SESSION_LAST_OPENED_KEY]: Date.now() },
```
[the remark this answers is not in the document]
MrOrz: Fixed

src/routes/_app/session.$sessionId.tsx
```diff
@@ -1,17 +1,27 @@
...
+  const queryClient = useQueryClient()
   const { messages, isStreaming, error, sendMessage, stopGeneration } = useChat(
     { sessionId },
   )
 
+  useEffect(() => {
+    markSessionOpened({ data: sessionId }).then(() => {
+      queryClient.invalidateQueries({ queryKey: ['sessions'] })
+    })
+  }, [sessionId, queryClient])
```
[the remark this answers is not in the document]
MrOrz: Makes sense. please also add comment indicating why `messages.length` is added here.

While `markSessionOpened` seems pretty frequent, it is the only reliable way to update the session open timestamp, given the fact that the user would close the page anytime.

However, `queryClient.invalidateQueries({ queryKey: ['sessions'] })` may trigger fetch requests too often. Consider update Tanstack query cache optimistically in `markSessionOpened` instead of invalidating the whole sessions cache.

src/components/Sidebar.tsx
```diff
@@ -121,7 +139,10 @@ export function Sidebar({ isOpen, onClose }: SidebarProps) {
   const currentSessionId = (params as Record<string, string | undefined>)
     .sessionId
 
-  const { data: sessions, isLoading } = useSessions()
+  const { data: rawSessions, isLoading } = useSessions()
+  const sessions = rawSessions
+    ?.slice()
+    .sort((a, b) => (b.lastEventTime ?? 0) - (a.lastEventTime ?? 0))
```
MrOrz: nit: I think we should directly sort the sessions in `sessions.functions.ts` instead.

src/lib/sessions.functions.ts
MrOrz: Please also setup lastEventTime and lastOpenTime here. New sessions should always have these states, otherwise the new session will display at very bottom and then lifted after the first response.

MrOrz: nit: maybe we should use the name `updateChatName` so that it can be as specific as the newly added `markSessionOpened`

MrOrz (commented):
@claude look at my comments and plan fixing them. you may ignore the nitpicking ones you think not important.

src/lib/sessions.functions.ts
MrOrz: fixed

MrOrz: fixed

src/components/Sidebar.tsx
```diff
@@ -121,7 +137,10 @@ export function Sidebar({ isOpen, onClose }: SidebarProps) {
   const currentSessionId = (params as Record<string, string | undefined>)
     .sessionId
 
-  const { data: sessions, isLoading } = useSessions()
+  const { data: rawSessions, isLoading } = useSessions()
+  const sessions = rawSessions
+    ?.slice()
+    .sort((a, b) => b.lastUpdateTime - a.lastUpdateTime)
```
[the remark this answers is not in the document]
MrOrz: moved

nonumpa (approved):
LGTM!

PR #51, opened 2026-05-09
feat: thread authenticated user id through chat, feedback, and ADK

### LINKED ISSUE #47: feat: 深度整合 userId
根據 [20260505 會議記錄](https://g0v.hackmd.io/4kQ9ihNOQfmtYvGcBkrKJg)，在登入機制 PR (https://github.com/cofacts/ai/pull/25) merge 後，需要進行以下實作：

- 將 user ID 放進 ADK endpoint，讓不同使用者的 side panel 可以不同
- 在 Langfuse feedback 中記錄 user id
- 確保未登入的使用者不能與 bot 聊天（可考慮不顯示聊天介面）

nonumpa:
Implements cofacts/ai#47.

## Summary

Thread the authenticated Cofacts user id end-to-end through chat, ADK, and Langfuse so per-user state (sessions, feedback, side panels) stays isolated. Authenticated ADK identity is resolved by verifying the BFF session JWT locally against rumors-api JWKS, avoiding per-request `GetUser` calls while preserving re-login behavior when the cookie expires. Replace per-region "logged-out" gating with a single layout-level landing.

### Per-user identity end-to-end
- `src/server/adkUser.ts`, `src/server/jwt.ts`, `src/server/jwks.ts` — resolve the current ADK `user_id` from the HttpOnly `cofacts_session` JWT. The token is verified locally with `jose.jwtVerify` against `COFACTS_API_URL/.well-known/jwks.json`, pins `RS256`, and uses the `sub` claim as the Cofacts user id.
- Auth failures use the shared `AUTH_EXPIRED` sentinel so TanStack server function callers can trigger the global re-login flow after serialization.
- `src/lib/chatSessions.functions.ts`, `src/routes/api/run-sse.ts` — drop the hard-coded `ADK_USER_ID`; resolve per-request.

### Auth expiry and re-login handling
- `src/lib/authExpired.ts`, `src/router.tsx`, `src/lib/auth.tsx` — centralize client-side 401/auth-expired handling. Query/mutation errors clear user-scoped caches and open `LoginModal` with the current pathname as the post-login redirect target.
- `src/lib/chatCache.ts`, `src/routes/api/run-sse.ts` — SSE requests also participate in the same auth-expired flow instead of only showing a generic stream error.

### Feedback persistence
- `src/server/feedbackScores.functions.ts` — new server functions read and write the current user's per-trace feedback score through the Langfuse REST API, filtered server-side by `trace.userId`.
- `FeedbackButtons` seeds local state from `getFeedbackForTrace` so thumbs-up/down survives reloads and tab switches.
- Feedback writes now go through `submitFeedbackForTrace`, re-resolving auth on every submission so expired cookies produce the same re-login flow as other authenticated writes.
- Cache key includes `userId`; `clearUserScopedCache` evicts `['feedback']` on logout.

### Logout cache hygiene
- `src/lib/auth.tsx` — `logout()` now removes user-scoped React Query caches (`['sessions']`, `['chat']`, `['feedback']`), not just `['me']`. These caches can use long lifetimes, so `removeQueries` is needed to prevent the previous user's data leaking to an anonymous viewer in the same tab.

### Logged-out landing (single switch)
- `src/routes/_app.tsx` — when `useAuth().user` is null, render only `<Header />` + `<LoggedOutLanding />` and skip the sidebar + `<Outlet />` entirely.
- New `LoggedOutLanding`, `WelcomeHero`, refined `LoginPrompt` (now requires explicit `message` prop).
- Removes scattered `user ?` branches in `ChatArea` and the landing route, and prevents authenticated-only fetches (sessions, chat history, feedback scores) from firing while logged out.

## Screenshots

Logged-out landing — desktop:

Logged-out landing (desktop)

Logged-out landing — mobile:

Logged-out landing (mobile)

## Known issue

- If a browser tab is left on `/session/:sessionId` and the user switches to another Cofacts account, the URL may point to a session owned by the previous ADK user. ADK correctly returns `404 Session not found`, but the UI currently surfaces it as a generic ADK communication error instead of redirecting home or explaining that the session does not belong to the current account. This is intentionally left for a follow-up.

## Notes

- No new env vars beyond the optional `LANGFUSE_*` server-side keys already documented in `.env.example` (used to enable feedback persistence; absent → feature degrades gracefully to "no persisted score").
- Cross-tab logout sync (tab A logs out → tab B still shows logged-in UI until next API call returns 401) is intentionally out of scope; matches current cofacts.tw behaviour.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,13 +5,9 @@ COFACTS_API_URL=https://dev-api.cofacts.tw
 # Server-side (optional) — ADK agent URL, defaults to http://localhost:8000
 # ADK_URL=http://localhost:8000
 
-# Browser-safe (public key only, VITE_ prefix exposes to Vite client bundle)
-VITE_LANGFUSE_PUBLIC_KEY=pk-lf-...
-VITE_LANGFUSE_BASE_URL=https://langfuse.cofacts.tw
-
 # Server-side (optional) — Langfuse keys for reading back user feedback scores.
-# Required to persist thumbs-up/down across reloads. Same key pair as the
-# browser-safe ones above; secret key MUST stay server-side.
+# Required to persist thumbs-up/down across reloads. Secret key MUST stay
+# server-side; all Langfuse traffic is proxied through the BFF.
 LANGFUSE_BASE_URL=https://langfuse.cofacts.tw
 LANGFUSE_PUBLIC_KEY=pk-lf-...
 LANGFUSE_SECRET_KEY=sk-lf-...
--- a/.env.example
+++ b/.env.example
@@ -8,3 +8,10 @@ COFACTS_API_URL=https://dev-api.cofacts.tw
 # Browser-safe (public key only, VITE_ prefix exposes to Vite client bundle)
 VITE_LANGFUSE_PUBLIC_KEY=pk-lf-...
 VITE_LANGFUSE_BASE_URL=https://langfuse.cofacts.tw
+
+# Server-side (optional) — Langfuse keys for reading back user feedback scores.
+# Required to persist thumbs-up/down across reloads. Same key pair as the
+# browser-safe ones above; secret key MUST stay server-side.
+LANGFUSE_BASE_URL=https://langfuse.cofacts.tw
+LANGFUSE_PUBLIC_KEY=pk-lf-...
+LANGFUSE_SECRET_KEY=sk-lf-...
--- a/src/components/FeedbackButtons.tsx
+++ b/src/components/FeedbackButtons.tsx
@@ -35,7 +35,6 @@ export function FeedbackButtons({ traceId }: FeedbackButtonsProps) {
           value: 0,
           dataType: 'NUMERIC',
           comment: '',
-          metadata: { userId: user?.id ?? null },
         })
       }
     } else {
@@ -48,7 +47,6 @@ export function FeedbackButtons({ traceId }: FeedbackButtonsProps) {
           name: 'user-thumbs',
           value: next,
           dataType: 'NUMERIC',
-          metadata: { userId: user?.id ?? null },
         })
       }
     }
@@ -64,7 +62,6 @@ export function FeedbackButtons({ traceId }: FeedbackButtonsProps) {
         value: feedbackGiven,
         dataType: 'NUMERIC',
         comment,
-        metadata: { userId: user?.id ?? null },
       })
     }
   }
--- a/src/components/FeedbackButtons.tsx
+++ b/src/components/FeedbackButtons.tsx
@@ -2,6 +2,7 @@ import { useState } from 'react'
 import { LangfuseWeb } from 'langfuse'
 import { Popover, PopoverContent, PopoverTrigger } from './ui/popover'
 import { FeedbackPopoverContent } from './FeedbackPopoverContent'
+import { useAuth } from '@/lib/auth'
 
 const langfuse = import.meta.env.VITE_LANGFUSE_PUBLIC_KEY
   ? new LangfuseWeb({
@@ -15,6 +16,7 @@ interface FeedbackButtonsProps {
 }
 
 export function FeedbackButtons({ traceId }: FeedbackButtonsProps) {
+  const { user } = useAuth()
   const [feedbackGiven, setFeedbackGiven] = useState<1 | -1 | null>(null)
   const [isPopoverOpen, setIsPopoverOpen] = useState(false)
 
@@ -33,6 +35,7 @@ export function FeedbackButtons({ traceId }: FeedbackButtonsProps) {
           value: 0,
           dataType: 'NUMERIC',
           comment: '',
+          metadata: { userId: user?.id ?? null },
         })
       }
     } else {
@@ -45,6 +48,7 @@ export function FeedbackButtons({ traceId }: FeedbackButtonsProps) {
           name: 'user-thumbs',
           value: next,
           dataType: 'NUMERIC',
+          metadata: { userId: user?.id ?? null },
         })
       }
     }
@@ -60,6 +64,7 @@ export function FeedbackButtons({ traceId }: FeedbackButtonsProps) {
         value: feedbackGiven,
         dataType: 'NUMERIC',
         comment,
+        metadata: { userId: user?.id ?? null },
       })
     }
   }
--- a/src/components/FeedbackButtons.tsx
+++ b/src/components/FeedbackButtons.tsx
@@ -1,17 +1,12 @@
 import { useEffect, useState } from 'react'
-import { useQuery } from '@tanstack/react-query'
-import { LangfuseWeb } from 'langfuse'
+import { useMutation, useQuery } from '@tanstack/react-query'
 import { Popover, PopoverContent, PopoverTrigger } from './ui/popover'
 import { FeedbackPopoverContent } from './FeedbackPopoverContent'
 import { useAuth } from '@/lib/auth'
-import { getFeedbackForTrace } from '@/server/feedbackScores.functions'
-
-const langfuse = import.meta.env.VITE_LANGFUSE_PUBLIC_KEY
-  ? new LangfuseWeb({
-      publicKey: import.meta.env.VITE_LANGFUSE_PUBLIC_KEY,
-      baseUrl: import.meta.env.VITE_LANGFUSE_BASE_URL,
-    })
-  : null
+import {
+  getFeedbackForTrace,
+  submitFeedbackForTrace,
+} from '@/server/feedbackScores.functions'
 
 interface FeedbackButtonsProps {
   traceId: string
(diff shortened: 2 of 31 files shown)


src/components/FeedbackButtons.tsx
```diff
@@ -35,6 +37,7 @@ export function FeedbackButtons({ traceId }: FeedbackButtonsProps) {
           value: 0,
           dataType: 'NUMERIC',
           comment: '',
+          metadata: { userId: user?.id ?? null },
```
MrOrz: I found that the userId is actually associated with trace ._.

Image

So we probably don't need to provide metadata here

.env.example
```diff
@@ -8,3 +8,10 @@ COFACTS_API_URL=https://dev-api.cofacts.tw
 # Browser-safe (public key only, VITE_ prefix exposes to Vite client bundle)
 VITE_LANGFUSE_PUBLIC_KEY=pk-lf-...
 VITE_LANGFUSE_BASE_URL=https://langfuse.cofacts.tw
```
MrOrz: Since we communicate to Langfuse in backend (BFF), we probably don't need these anymore

src/server/feedbackScores.functions.ts
```diff
@@ -0,0 +1,153 @@
...
+  }
+}
+
+export const submitFeedbackForTrace = createServerFn({ method: 'POST' })
+  .inputValidator((raw: unknown): SubmitFeedbackInput => {
+    if (!isSubmitFeedbackInput(raw)) {
+      throw new Error('Invalid feedback submission payload')
+    }
+    return raw
+  })
```
MrOrz: nit: I think we don't need runtime validator here, because the type is already protected by Typescript.
We can leave validators for something that cannot be protected using static type analysis.

src/server/me.functions.ts
```diff
@@ -1,14 +1,19 @@
 // TanStack Start server function exposing the current logged-in user to the
 // client. Reads the HttpOnly cofacts_session cookie via h3's getCookie and
-// dispatches the GetUser GraphQL query through cofactsExec. Always resolves —
-// never throws — so loaders and effects can call it without try/catch
-// boilerplate; cofactsExec throws are caught here and mapped to null.
+// dispatches the GetUser GraphQL query through cofactsExec. If GetUser fails
+// (upstream down, transient error) but a valid session cookie exists, falls
+// back to a minimal user derived from the JWT sub claim so the app keeps
```
MrOrz: I have read through the commit message of 3815ca879da4b3282f447c0cee890e51433f6d1b but still don't quite get it.

As far as I know from the commit message, a hiccup in the API will cause `user` being falsy and thus kicks the user to the landing page due to the `!user` gate in `_app.tsx`.

However, in `auth.tsx` user is loaded using this:
```typescript
  const { data: user, isFetching } = useQuery<CofactsUser | null>({
    queryKey: ME_QUERY_KEY,
    queryFn: () => getCurrentUserServerFn(),
    initialData: serverLoadedUser ?? null,
    staleTime: Infinity,
  })
```

Since the `staleTime` is infinity, `user` is never invalidated unless `clearUserScopedCache` is called. So `!user` should not be triggered once the page is loaded. Even if the Cofacts API is really down, I don't see if `clearUserScopedCache()` would be called to clean up `user` in this case.

How could the a transient API server error send the user back to not-logged-in page?

src/lib/auth.tsx
```diff
@@ -48,6 +70,17 @@ export function AuthProvider({
...
 
+  // Open LoginModal whenever any client-side call detects 401. The redirect
+  // target is the current pathname so the user lands back where they were
+  // after re-authenticating.
+  useEffect(() => {
+    const onAuthExpired = () => {
+      setPendingRedirect(router.state.location.pathname)
+    }
+    window.addEventListener(AUTH_EXPIRED_EVENT, onAuthExpired)
+    return () => window.removeEventListener(AUTH_EXPIRED_EVENT, onAuthExpired)
```
MrOrz, line 82: Since `handleAuthExpired` will invalidate `user`, I think we can also detect user being emptied instead of relying on a custom `AUTH_EXPIRED_EVENT` event.

If leveraging global event is still preferred, I would suggest we double down and rely on the mechanism more:

1. We can move the call to `clearUserScopedCache` into `onAuthExpired` here, so that `handleAuthExpired` only need to do `window.dispatchEvent()`, and we no longer need to pass `queryClient` everywhere `handleAuthExpired` is called.
2. Let's also introduce [BroadcastChannel](https://developer.mozilla.org/en-US/blog/exploring-the-broadcast-channel-api-for-cross-tab-communication/) so that `handleAuthExpired` fires custom event (via `window.dispatchEvent()`) and also sends the logout event through a `BroadcastChannel` to all browser tabs that is opening the same page. In the `useEffect` here we listen to both the window (for own event) and the `BroadcastChannel` to open the LoginModal.

src/routes/_app/index.tsx
```diff
@@ -36,6 +38,11 @@ function LandingPage() {
           },
         })
       } catch (err) {
+        if (isAuthExpiredError(err)) {
+          handleAuthExpired(queryClient)
+          setIsLoading(false)
+          return
+        }
```
MrOrz: nit: I was not familiar with `useMutation` when I write `handleSend`.  Now I see that we have `loading`, `error` states for this, I am thinking we should probably rewrite this using `useMutation`. This should also save us from handling `AuthExpiredError` manually here, since it is handled globally when we instantiate `QueryClient` in `router.tsx`.

src/server/me.functions.ts
nonumpa: We have **two paths consuming "who is the user", and only one of them currently treats JWT as the source of truth**:

1. **ADK path** (server-side, fixed in this PR): reads `cofacts_session`, verifies the JWT, takes `sub` as `userId`. No GetUser dependency.
2. **UI auth path** (client gate in `_app.tsx`): reads [`useAuth().user`](https://github.com/cofacts/ai/pull/51/changes#diff-45f022562523df1ac8dea1cab1e9ddeedab9407f0949d955e5c4d568f5a75c9fR25), which comes from `getCurrentUserServerFn()` — and that fn still calls `GetUser` against rumors-api to fetch the full profile (id + name + avatarUrl + avatarType + avatarData).

`_app.tsx`'s `!user` gate therefore conflates two different questions:

- "Is this session authenticated?" — should depend only on JWT validity
- "What's the profile for the authenticated user?" — depends on GetUser, which can fail or rate-limit

When those two get tangled, GetUser availability decides whether the app renders the main shell or `LoggedOutLanding`. Combined with `AuthProvider`'s `initialData` + `staleTime: Infinity`, an SSR `null` sticks in the cache and the client never auto-refetches — hence the "keep refreshing and eventually get bounced to landing" symptom.

The current state in this PR:

| GetUser result | JWT state | Returned | Intended meaning |
|---|---|---|---|
| throws | valid | fallback minimal user | authenticated, profile temporarily missing |
| returns `{ GetUser: null }` (200) | valid | **`null`** | authenticated, profile temporarily missing — but we return logged-out |
| success | valid | full user | authenticated |
| any | invalid / missing | `null` | logged out |

[`3815ca8`](https://github.com/cofacts/ai/pull/51/changes/3815ca879da4b3282f447c0cee890e51433f6d1b) covers the `throws` row only. The `200 + GetUser: null` row still leaks through and gets locked in by `staleTime: Infinity`.

**Plan**: keep this PR's scope as-is and follow up with a refactor that splits auth state from profile state, so the UI path aligns with the ADK path on JWT-as-source-of-truth. Sketch:

1. New server fn `getAuthStatusServerFn`: verifies the JWT only, returns `{ authenticated, userId }`. No rumors-api call.
2. `getCurrentUserServerFn` keeps returning `CofactsUser | null` but is purely cosmetic — name/avatar fetch, allowed to fail, allowed to retry.
3. `AuthProvider` exposes two queries:
   - `['auth']` — authoritative gate, driven by JWT
   - `['me']` — profile, independent retry / shorter staleTime / safe to be `null`
4. `_app.tsx` gate switches from `!user` to `!auth.authenticated`.
5. Drop the JWT-sub fallback in `getCurrentUserServerFn` once `['me']` no longer doubles as the auth signal.

That keeps "is the user logged in" as a local, synchronous check that doesn't depend on the upstream API, and degrades `GetUser` failures into a missing-name placeholder instead of a forced logout.

nonumpa: Plan for a separate PR with this shape — listing files for visibility:

**New**
- `src/server/auth.status.functions.ts` — `getAuthStatusServerFn`, JWT-only. Returns `{ authenticated, userId }`. No rumors-api call.

**Modified**
- `src/server/me.functions.ts` — drop the JWT-sub fallback. Returns `CofactsUser | null` purely as profile data; `null` no longer means "logged out", just "profile unavailable".
- `src/lib/auth.tsx` — `AuthProvider` runs two queries:
  - `['auth']` ← `getAuthStatusServerFn`, authoritative gate
  - `['me']` ← `getCurrentUserServerFn`, cosmetic profile (shorter staleTime, independent retry)
  - `AuthState` exposes `authenticated`, `userId`, `user`. `clearUserScopedCache` clears both keys.
- `src/routes/__root.tsx` — loader fetches both server fns in parallel, passes `serverLoadedAuth` + `serverLoadedUser` to `AuthProvider`.
- `src/routes/_app.tsx` — gate switches from `!user` to `!authenticated`.

**Unchanged (consumers of profile)**
- `src/components/Header.tsx`, `src/components/FeedbackButtons.tsx` — keep reading `user` for display; render placeholders when `null` while authenticated.

src/routes/_app/index.tsx
MrOrz: So elegant after [rewrite](https://github.com/cofacts/ai/pull/51/commits/f7edefe440549e6a53217922e363235b9a6f28d9)! Thanks!

src/server/me.functions.ts
MrOrz: Agree with the refactor direction 👍🏻

MrOrz (approved):

src/lib/auth.tsx
nonumpa: Left BroadcastChannel out of this round; will open a separate issue for
cross-tab logout sync.

PR #67, opened 2026-05-23
Fix Langfuse score read fields

nonumpa:
## What changed

- Change the Langfuse score read query from `fields=scores,trace` to `fields=score,trace`.
- Update the feedback score unit test to match the actual Langfuse v2 API parameter.

## Why

Feedback submission now works, but reloading the page did not restore the previous thumbs-up/down state. The read side was calling `/api/public/v2/scores` with `fields=scores,trace`, which Langfuse rejects with `400 Bad Request` and the message `Scores needs to be selected always.` The accepted field group is singular: `score`.

## Validation

- `pnpm vitest run src/server/__tests__/feedbackScores.test.ts` passes 12 tests.
- Queried the real Langfuse API for the PR preview trace with `fields=score,trace`; it returned `200` and the expected `user-thumbs` score.


MrOrz (approved):
Image 
Verified that it works, thanks!

PR #76, opened 2026-05-30
Pass auth token via Authorization header + ContextVar instead of session stateDelta

MrOrz:
## Problem

The previous approach in `run-sse.ts` sent the Cofacts JWT as `stateDelta: { 'temp:cofacts_token': token }`. This never worked: ADK's `BaseSessionService.append_event` calls `_trim_temp_delta_state` which strips all `temp:` prefixed keys from the event before they reach `session.state`, so `tool_context.state.get("temp:cofacts_token")` always returned `None`.

## Root cause (traced through ADK source)

`temp:` state is designed to be set by **Python callbacks within an invocation** via `callback_context.state["temp:foo"] = bar` (which mutates `session.state` dict directly). When passed via the HTTP `stateDelta` field, it goes through `EventActions` → `append_event` → `_trim_temp_delta_state` → stripped, never lands in `session.state`.

## Solution

Use a `ContextVar` + FastAPI middleware instead:

1. **`adk/cofacts_ai/auth_context.py`** — defines `cofacts_token_var: ContextVar[Optional[str]]`
2. **`adk/main.py`** — replaces `adk api_server` CLI; adds an HTTP middleware that reads `Authorization: Bearer <token>` and sets the ContextVar
3. **`adk/cofacts_ai/tools.py`** — tools call `cofacts_token_var.get()` directly
4. **`src/routes/api/run-sse.ts`** — sends `Authorization: Bearer <token>` header instead of `stateDelta`
5. **`adk/Dockerfile`** — CMD changed to `python main.py`

### Why ContextVar propagates to tools

FastAPI middleware → `call_next` → Starlette's `BaseHTTPMiddleware` uses `anyio.create_task_group().start_soon(coro)`. Python asyncio automatically copies the current `contextvars` context into new tasks, so `cofacts_token_var` set in the middleware is visible everywhere in the request handling chain — including deep inside `runner.run_async` → tool functions.

No session state is touched; the token never appears in `list_sessions` responses.

## Test plan

- [ ] Start ADK server with `python main.py`
- [ ] Make a `/run_sse` request with a valid session cookie; confirm tools receive the token
- [ ] Verify Cofacts GraphQL calls with auth succeed (e.g. `submit_cofacts_reply`)
- [ ] Verify unauthenticated requests (no cookie) still work — tools receive `None` and make unauthenticated GraphQL calls as before

https://claude.ai/code/session_01Ke7rQW8p5rHtUk5Nm4HDFY

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ke7rQW8p5rHtUk5Nm4HDFY)_

### DIFF
--- a/adk/Dockerfile
+++ b/adk/Dockerfile
@@ -13,4 +13,4 @@ COPY --from=builder /app/.venv /app/.venv
 ENV PATH="/app/.venv/bin:$PATH"
 EXPOSE 8000
 ENV DATABASE_URL=sqlite+aiosqlite:////tmp/sessions.db
-CMD exec adk api_server --port 8000 --host 0.0.0.0 --session_service_uri=$DATABASE_URL .
+CMD exec python main.py
--- a/adk/cofacts_ai/auth_context.py
+++ b/adk/cofacts_ai/auth_context.py
@@ -1,0 +1,8 @@
+from contextvars import ContextVar
+from typing import Optional
+
+# Populated by the FastAPI middleware in main.py for each /run_sse request.
+# Tools read this directly instead of going through session state.
+cofacts_token_var: ContextVar[Optional[str]] = ContextVar(
+    "cofacts_token", default=None
+)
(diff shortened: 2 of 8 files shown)


MrOrz:
We are getting original video URL from `get_single_cofacts_article`, seems to be working now
image

PR #79, opened 2026-06-03
Register Cofacts media via Gemini Files API instead of signed URL

MrOrz:
## Background

Investigating Langfuse session `71212151-5de2-43ed-b543-4b77a2163027` (a user submitting a Cofacts **video** article) confirmed the writer **does** read the video (verified: `usage_metadata` shows `VIDEO` modality, 7098 tokens). But two problems surfaced:

1. **Redundant verifier call.** The writer also forwarded the article's *signed* GCS URL (as plain text) to the `verifier` sub-agent. The verifier has no `inject_article_attachment` and its `inject_youtube_filedata` only matches YouTube URLs, so it can't watch the video; `url_context` can't read a raw signed file URL either → verifier returned `{"result": "<not specified>"}` (empty), stalling the fact-check flow.
2. **Signed-URL expiry.** The signed `attachmentUrl` expires (~37h in that session). Since `inject_article_attachment` re-injects the captured URL from conversation history every turn, reopening a conversation after expiry would make Gemini's GCS fetch fail.

## Change

Switch the media-attachment path from a directly-injected signed HTTPS URL to **GCS URI registration** with the Gemini Files API:

- **`after_tool`** rewrites `get_single_cofacts_article`'s `attachmentUrl` from the signed HTTPS URL to a `gs://` URI. The writer (and conversation history) now only see a non-fetchable `gs://` reference, so it can't forward a usable signed URL to the verifier.
- **`inject_article_attachment`** registers that `gs://` object via `client.files.register_files(...)` and injects the returned `files/...` handle as `file_data`. The `gs://` URI never expires and is re-read from history every turn, so follow-up turns keep working without re-calling the tool. Registration grants access for up to 30 days and Gemini fetches via the Service Agent's bucket permission. On registration failure it logs and skips injection (no crash).
- Adds `google-auth` as an explicit dependency; ADC is used for the registration credentials (`devstorage.read_only` scope).

## ⚠️ Deployment prerequisite (IAM)

The bucket `cofacts-media-collection` lives in the Cofacts GCP project (`industrious-eye-145611`). Before this works in each environment, grant `roles/storage.objectViewer` on the bucket to **both**:

1. **The app service account** (used at registration time):
   - **Cloud Run (staging/preview):** the runtime `SERVICE_ACCOUNT_EMAIL` (ADC via metadata, no key file).
   - **docker-compose (production, outside GCP):** a SA whose JSON key is mounted and pointed to by `GOOGLE_APPLICATION_CREDENTIALS` (see `.env.example`).
2. **The Gemini API Service Agent** (used at fetch time), of the `GOOGLE_API_KEY`'s project: `gcloud beta services identity create --service=generativelanguage.googleapis.com` then grant it viewer. Cross-project grant is needed if that project ≠ `industrious-eye-145611`.

Until IAM is in place, media injection is skipped (logged), so existing text behavior is unaffected.

## Verification

- Unit: `_signed_url_to_gs` correctly converts path-style and virtual-hosted GCS URLs and drops the query string (verified locally).
- `register_files` exists on both sync and async clients in the pinned `google-genai` 1.66.0 (added in 1.57.0).
- Once IAM is granted: re-run the video article (`prqoNp0B-5SQhC6hlKzX`) and confirm the injected `file_data.file_uri` is a `files/...` handle, the writer-visible tool response contains only `gs://`, `VIDEO` modality tokens are present, and no empty `verifier` call occurs.

https://claude.ai/code/session_01U6uR2yvsAvZsRto3s11G5P

---
_Generated by [Claude Code](https://claude.ai/code/session_01U6uR2yvsAvZsRto3s11G5P)_


MrOrz:
Closing without merging — abandoning the Files-API GCS-registration approach.

**Conclusion / reasoning:**

1. **The permissions cost isn't worth it.** GCS `register_files` needs broad bucket access granted to *two* identities (the app SA *and* the Gemini API Service Agent, cross-project). That's a heavier grant than just letting the app's own service account re-sign a signed URL — at which point we may as well keep signing URLs ourselves.

2. **GCS registration doesn't work on the model we want to use.** The registered `files/...` `fileUri` makes the current `gemini-3-flash-preview` writer emit empty/garbage output (see #80), and we don't want to move to a more expensive model just to make registration work.

3. **Signed URLs are the known-good, simpler path.** The signed-URL delivery already produces correct analysis on the same model. So we'll stay on signed URLs and only re-sign — using the app's own service account — when a URL is confirmed expired. Less code, fewer IAM prerequisites, no dependency on model-side `fileUri` handling.

The dependent PR #80 (writer → `gemini-2.5-flash`) is moot under this decision and is already closed. The branch remains if we ever want to revisit registration.

---

MrOrz:
**Evidence — Langfuse traces where GCS-registration delivery fails**

Both runs are the same video article (`prqoNp0B-5SQhC6hlKzX`) with the Files-API `files/...` handle injected. The writer's 2nd LLM call consumes `VIDEO 7098` tokens (the Service-Agent fetch succeeds) but the model emits no thinking and degenerate output:

- Run 1 — output `≧00:00` (6 junk tokens): https://langfuse.cofacts.tw/project/cmm0emerr0001qi07eugd0760/traces/f42f74f3d0005ceee1145f3364954057
- Run 2 — empty output: https://langfuse.cofacts.tw/project/cmm0emerr0001qi07eugd0760/traces/19608d44d7b23c611d41dbb2b1701a89

For contrast, the signed-URL delivery produced a correct analysis on the same model / same 7098 video tokens (the working baseline from #80's investigation).

_(Traces are private — Langfuse login required.)_

---

PR #82, opened 2026-06-06
feat(adk): deliver Cofacts media to Gemini via Vertex AI + native gs://

MrOrz:
## Summary

Stacked on #72. Switches the agent from the Gemini **Developer API** (API key) to **Vertex AI**, and delivers article media to Gemini as native **`gs://`** URIs instead of signed HTTPS URLs. This deletes the signed-URL re-signing machinery added in #81.

### Update — writer works from transcript, verifier is the sole video/audio watcher (commit `02b63f3`)

While verifying the media path, traces showed that handing the **writer/orchestrator** temporal media (VIDEO/AUDIO) destabilizes it: it narrates/confabulates the playback and gets captured by the audio narrative instead of orchestrating. This commit reverses #72's "writer perceives all media directly" premise **for temporal media only**:

- `inject_article_attachment` now injects **IMAGE only** (`_WRITER_INJECTED_TYPES`). A still image has no narration to parrot and the writer genuinely needs to see it (manipulated charts/photos).
- For **VIDEO/AUDIO**, the writer works from the **transcript Cofacts already extracts into the article text** for spoken claims, and **delegates watching/listening to the verifier** — the sole video-watcher (it already receives the `gs://` media as `FileData`).
- The **verifier instruction** now requires an **exhaustive, atomic, numbered claim inventory** covering both the spoken and visual layers, and supports **targeted re-watch** follow-ups from the writer.
- The writer Step-2 instruction reflects the transcript-first / delegate-the-visual-layer split.

> Note: this reverses part of #72's thesis for video/audio. Landed here (not on #72) per discussion. Pre-existing ruff errors on this branch (unused `submit_cofacts_reply` import) are unrelated to this change and left untouched.

### Why (Vertex + gs://)

Langfuse traces showed the signed-URL transport **already worked** — the video is fetched and tokenized (a constant `VIDEO 7098` tokens) on both correct and broken runs. So media delivery was never the issue; the signing/re-signing apparatus was avoidable complexity. On Vertex, `file_data.file_uri` accepts `gs://` natively (HTTP URLs are capped at ~15MB, which our ~80MB video exceeds), so `gs://` is *required* — and it lets us drop the on-demand re-signing entirely. rumors-api already runs on Vertex, so the project can read `gs://cofacts-media-collection`.

### Changes

- **Vertex env switch** (no model-instantiation code change — ADK builds its client from env): `GOOGLE_GENAI_USE_VERTEXAI=true`, `GOOGLE_CLOUD_PROJECT`, `GOOGLE_CLOUD_LOCATION=global`, ADC auth.
  - `adk/cofacts_ai/.env.example`, `service.template.yaml`, `.github/workflows/deploy.yml` swap `GOOGLE_API_KEY` for the Vertex vars. `GOOGLE_CLOUD_PROJECT` defaults to the deploy project (`GC_PROJECT_ID`); override with a `GOOGLE_CLOUD_PROJECT` secret.
- **`gcs_signing.py`**: gut the re-sign apparatus (credentials cache, expiry parsing, `_resign_gcs_blob`, `_refresh_attachment_url`) and the `google-cloud-storage` / `google-auth` imports. Add `signed_url_to_gs()`; `inject_article_attachment` now injects the **`gs://`** form.
- **Verifier now actually watches the video.** Previously the writer forwarded a signed URL in plain text and the verifier — which only has `url_context` — couldn&#39;t read a raw storage object, so it hallucinated the video content. Now:
  - `after_tool` rewrites the writer-visible `attachmentUrl` to `gs://`, so the writer only ever forwards (and the verifier only ever receives) a `gs://` URL — never a signed one.
  - a new `inject_cofacts_media_filedata` before-model callback on the verifier injects that `gs://` as `FileData` so Gemini watches it directly.
  - the verifier instruction excludes Cofacts media URLs from `url_context`.
- **Writer/verifier media split (commit `02b63f3`, see Update above):** writer gets IMAGE only; VIDEO/AUDIO handled by transcript + verifier claim inventory.
- Drop the `google-cloud-storage` direct dependency; regenerate `uv.lock`.

### Models unchanged in this pass

Kept on `gemini-3-flash-preview`. The previously-observed collapse/hallucination (empty `uid:…` output, `pid=…size=` handle leak, 500s, describing the porcelain video as &#34;eID/TSMC&#34; then &#34;phone-case&#34;) may be **Developer-API-specific**, so this PR uses Vertex itself as the test. Switching the video-consuming agents (writer + verifier) to `gemini-2.5-flash` (with `thinking_budget=-1`, since 2.5 rejects `thinking_level`) is held as a **contingency** if the issue persists on Vertex.

## Verification

Compile + import + functional unit checks pass locally (URL→gs:// conversion path-style/vhost/passthrough, MIME inference, regex matching, `after_tool` rewrite, writer IMAGE-only injection + verifier injection, idempotency). **Still to run on a Vertex-configured deployment** (needs `GOOGLE_CLOUD_PROJECT` + ADC + bucket read):

- [ ] Confirm Vertex is active (traces hit `*-aiplatform.googleapis.com`).
- [ ] Re-run article `prqoNp0B-5SQhC6hlKzX` (the porcelain video) 2–3×; the **verifier** describes the porcelain video (not eID/phone-case) and returns an atomic claim inventory, the **writer** composes from the verifier's inventory + the article transcript (writer no longer receives the video FileData), VIDEO tokens consumed on the verifier, no `uid:`/`pid=` leak / runaway loop / 500, consistent across runs. If `gemini-3-flash-preview` still misbehaves → apply the `gemini-2.5-flash` contingency.
- [ ] Smoke-test an IMAGE article: the writer still receives the image as `FileData` and analyzes it directly (image injection retained).
- [ ] `url_context` still populates sources on Vertex for a normal web URL (and the `gs://` media URL is absent from sources / triggers no url_context error).
- [ ] `google_search` (investigator) returns grounded sources on Vertex.
- [ ] All current model IDs resolve on Vertex `global` (`gemini-3-flash-preview`, `gemini-3.1-flash-lite`); fall back any that 404 to `gemini-2.5-flash`.
- [ ] Runtime SA can read `gs://cofacts-media-collection` and use Vertex.

https://claude.ai/code/session_015eRNbvR41ScL8uRfYDD8XX

---
_Generated by [Claude Code](https://claude.ai/code/session_015eRNbvR41ScL8uRfYDD8XX)_


.github/workflows/deploy.yml
```diff
@@ -128,7 +128,10 @@ jobs:
 
       - name: Generate Service YAML
         env:
-          GOOGLE_API_KEY: ${{ secrets.GOOGLE_API_KEY }}
+          # Vertex AI project for Gemini. Defaults to the deploy project; set a
+          # GOOGLE_CLOUD_PROJECT secret to override (it must have Vertex AI access
+          # and read access to the cofacts-media-collection bucket).
+          GOOGLE_CLOUD_PROJECT: ${{ secrets.GOOGLE_CLOUD_PROJECT || secrets.GC_PROJECT_ID }}
```
MrOrz: Let's use GC_PROJECT_ID throughout the project as it is more widely used in Cofacts project

adk/cofacts_ai/media_filedata.py
MrOrz, line 1: if the file does not do signing anymore, please find a better name for the file

.github/workflows/deploy.yml
MrOrz: Done — dropped the separate `GOOGLE_CLOUD_PROJECT` secret/override and now source the Vertex project from `GC_PROJECT_ID` (in `service.template.yaml` it's `value: "${GC_PROJECT_ID}"`). The container env var *name* stays `GOOGLE_CLOUD_PROJECT` because the google-genai Vertex SDK reads that specific variable.

---

adk/cofacts_ai/media_filedata.py
MrOrz: Renamed to `media_filedata.py` — it now converts media references to `gs://` URIs and injects them as `FileData` parts rather than signing. Import in `agent.py` updated accordingly.

---

PR #95, opened 2026-06-24
fix: add administration:write permission to preview-cleanup workflow

MrOrz:
Fixes #45

## Root cause

The `deleteAnEnvironment` GitHub API endpoint requires `administration=write` permission (confirmed from the 403 response header: `x-accepted-github-permissions: administration=write`). The `preview-cleanup.yml` workflow was missing this permission, causing all cleanup runs to fail with:

```
HttpError: Resource not accessible by integration (403)
```

## Fix

Added `administration: write` to the `permissions` block in `.github/workflows/preview-cleanup.yml`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UE511cMg9nN3LneAcNFLP7)_

PR #96, opened 2026-06-24
fix: replace deleteAnEnvironment with per-deployment cleanup

### LINKED ISSUE #45: Preview environment cleanup workflow does not work
Image

MrOrz:
Fixes #45, supersedes #95

## Problem

`administration` is not a valid GitHub Actions `permissions` scope, so the workflow file was invalid and no jobs ran.

The `deleteAnEnvironment` API requires admin-level access that cannot be granted via `GITHUB_TOKEN` permissions.

## Fix

Instead of deleting the environment directly, iterate over all deployments in the PR's environment, mark each as `inactive`, then delete them. This only requires `deployments: write` which was already present.

When all deployments in a transient environment are removed, GitHub cleans up the environment automatically.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UE511cMg9nN3LneAcNFLP7)_

### DIFF
--- a/.github/workflows/preview-cleanup.yml
+++ b/.github/workflows/preview-cleanup.yml
@@ -8,7 +8,6 @@ permissions:
   contents: read
   id-token: write
   deployments: write
-  administration: write
 
 jobs:
   cleanup:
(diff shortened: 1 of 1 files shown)


MrOrz:
## Code Review

**(none)** — no correctness bugs found.

The approach is sound: list deployments → mark each `inactive` → delete each. One minor note: `per_page: 100` will silently skip any deployments beyond the first 100, but for a PR preview environment this is unlikely to matter in practice.

---

PR #102, opened 2026-07-07
feat(session): auto-generate session title after first turn

### LINKED ISSUE #87: 自動為 session 命名
在 session 第一個 turn 結束時，使用 LLM 根據使用者的輸入與第一個 turn 內的結果，自動生成一個簡潔且有意義的標題。

參考會議記錄: https://g0v.hackmd.io/@cofacts/meetings/YF4cpCN2Q5W4llrFZoAf5A

SametAtas:
Closes #87.

At the end of a session's first turn, an LLM generates a concise title from the
user's first message and the writer's first result, replacing the placeholder
title (the raw message truncated to 40 chars) that the frontend sets at session
creation.

## Design
- New after_agent_callback `generate_session_title` in `adk/cofacts_ai/session_title.py`,
  registered alongside `update_last_event_time` on `ai_writer`.
- Runs on the first turn only (exactly one user event), so it never regenerates on
  later turns and never overwrites a title the user edited in the sidebar.
- Uses the cheap tier `gemini-3.1-flash-lite` via ADC/Vertex, thinking level MINIMAL,
  max_output_tokens 128, with a 10s timeout. Any failure is caught and the placeholder
  title is kept, so a title error never breaks the agent turn.
- Reuses a single lazily-initialized module-scoped genai client rather than constructing
  one per turn, mirroring the ADC-caching pattern adopted in #99.
- Stores the normalized title in session state under `title` — the key the frontend
  already reads (`listSessions`) and renders in the sidebar, so no frontend change is needed.

## Verification
Verified offline (mocking the genai client) across: first-turn title generation and
normalization, second-turn no-op (LLM not called), LLM failure leaving the title
unchanged, and a whitespace-only result not being written. Deterministic on repeat.
- `python -m unittest cofacts_ai.session_title_test` -> 4 passed, 0 failed
- `python -m py_compile` on changed files -> clean

The live integration test (an actual title generated end-to-end on Vertex) is still
pending, so I am keeping this as a Draft for now.

### DIFF
--- a/adk/cofacts_ai/session_title.py
+++ b/adk/cofacts_ai/session_title.py
@@ -111,10 +111,31 @@ def _truncate_prompt_text(text: str) -> str:
     return text[:PROMPT_TEXT_LIMIT].rstrip()
 
 
+# Quote characters the model sometimes wraps the whole title in. Stripped only as a
+# matched pair, so a title that legitimately contains a quote (\u4ed6\u8aaa\u300c\u9019\u662f\u5047\u7684\u300d) keeps it.
+_QUOTE_PAIRS = {
+    '"': '"',
+    "'": "'",
+    "`": "`",
+    "\u201c": "\u201d",
+    "\u2018": "\u2019",
+    "\u300c": "\u300d",
+    "\u300e": "\u300f",
+}
+
+
+def _strip_wrapping_quotes(title: str) -> str:
+    while len(title) >= 2:
+        closing = _QUOTE_PAIRS.get(title[0])
+        if closing is None or not title.endswith(closing):
+            break
+        title = title[1:-1].strip()
+    return title
+
+
 def _normalize_title(title: str) -> str:
     title = " ".join(title.split()).strip()
-    title = title.strip("\"'`\u201c\u201d\u2018\u2019\u300c\u300d\u300e\u300f")
-    title = title.strip()
+    title = _strip_wrapping_quotes(title)
     if len(title) > MAX_TITLE_LENGTH:
         title = title[:MAX_TITLE_LENGTH].rstrip()
     return title
--- a/adk/cofacts_ai/session_title.py
+++ b/adk/cofacts_ai/session_title.py
@@ -1,0 +1,120 @@
+import asyncio
+import logging
+
+from google import genai
+from google.genai import types as genai_types
+
+logger = logging.getLogger(__name__)
+
+TITLE_STATE_KEY = "title"
+TITLE_MODEL = "gemini-3.1-flash-lite"
+MAX_TITLE_LENGTH = 60
+PROMPT_TEXT_LIMIT = 4000
+TITLE_TIMEOUT_SECONDS = 10
+
+# Reuse a single genai client across sessions instead of constructing one (and
+# re-resolving ADC) on every first turn. Lazily initialized so importing this
+# module needs no credentials.
+_client = None
+
+
+def _get_client():
+    global _client
+    if _client is None:
+        _client = genai.Client()
+    return _client
+
+
+async def generate_session_title(callback_context) -> None:
+    """Generates a session title after the first writer turn."""
+    if _count_user_events(callback_context.session.events) != 1:
+        return None
+
+    user_text = _content_text(callback_context.user_content)
+    result_text = _last_writer_text(callback_context.session.events)
+    if not user_text and not result_text:
+        return None
+
+    try:
+        raw_title = await asyncio.wait_for(
+            _generate_title(user_text, result_text),
+            timeout=TITLE_TIMEOUT_SECONDS,
+        )
+    except Exception:
+        logger.exception("Failed to generate session title")
+        return None
+
+    title = _normalize_title(raw_title)
+    if title:
+        callback_context.state[TITLE_STATE_KEY] = title
+    return None
+
+
+def _count_user_events(events) -> int:
+    return sum(1 for event in events if event.author == "user")
+
+
+def _last_writer_text(events) -> str:
+    for event in reversed(events):
+        if event.author != "writer":
+            continue
+        text = _content_text(event.content)
+        if text:
+            return text
+    return ""
+
+
+def _content_text(content) -> str:
+    if not content or not content.parts:
+        return ""
+    texts = []
+    for part in content.parts:
+        text = getattr(part, "text", None)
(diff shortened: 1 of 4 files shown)


adk/cofacts_ai/session_title_test.py
```diff
@@ -0,0 +1,113 @@
...
+            [
+                _event("user", "請查證"),
+                _event("writer", "查證結果"),
+            ],
+            user_text="請查證",
+            title="請查證",
+        )
+
+        with patch("cofacts_ai.session_title.genai.Client", return_value=client):
+            await generate_session_title(context)
```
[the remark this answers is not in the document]
SametAtas: Thanks for the review. Addressed in the latest push:

- Module-scoped, lazily-initialized genai client (reused across turns, not
  reconstructed per call) — this was already in the current head; the earlier
  comment reviewed the pre-rebase commit.
- Tests now patch `_get_client` to match the caching design.

On the defensive `getattr` suggestions for `event.author` / `content` / `parts`:
I left those as direct attribute access. These are ADK `Event`/`Content` pydantic
models where the fields are always present, so the access can't raise
`AttributeError` in practice; guarding them would add handling for states that
can't occur.

MrOrz:
Tested locally, worked like a charm 🎉 


https://github.com/user-attachments/assets/3fb83a57-d364-4d3b-917c-4e3f8b5039ec

adk/cofacts_ai/session_title.py
```diff
@@ -0,0 +1,120 @@
...
+            _generate_title(user_text, result_text),
+            timeout=TITLE_TIMEOUT_SECONDS,
+        )
+    except Exception:
+        logger.exception("Failed to generate session title")
+        return None
+
+    title = _normalize_title(raw_title)
+    if title:
+        callback_context.state[TITLE_STATE_KEY] = title
```
MrOrz, line 49: **correctness:** Auto-generated title can silently overwrite a user's manual session rename made during the first turn.

The Sidebar rename button is not gated by `isStreaming` and is clickable as soon as the answer text is visible (well before the up-to-10s-delayed `after_agent_callback` resolves). If the user reads the finished answer and immediately renames the session via the sidebar, that PATCH's `state_delta` write can land before this delayed `callback_context.state[TITLE_STATE_KEY] = title` write. Both go through the same last-write-wins state-merge code in the session service, so the auto-generated title can revert the user's manual rename with no indication why.

adk/cofacts_ai/session_title.py
```diff
@@ -0,0 +1,120 @@
...
+
+def _truncate_prompt_text(text: str) -> str:
+    if len(text) <= PROMPT_TEXT_LIMIT:
+        return text
+    return text[:PROMPT_TEXT_LIMIT].rstrip()
+
+
+def _normalize_title(title: str) -> str:
+    title = " ".join(title.split()).strip()
+    title = title.strip("\"'`\u201c\u201d\u2018\u2019\u300c\u300d\u300e\u300f")
```
MrOrz: **correctness:** `_normalize_title` strips quote/bracket characters as an independent-per-edge character set, not a matched pair, so a title with a bracket only at one edge is left with a mismatched bracket.

Repro: `"「南部大停電」證實為假消息".strip("\"'`“”‘’「」『』")` returns `'南部大停電」證實為假消息'` — the leading 「 is stripped since it's at the edge, but the trailing 」 (mid-string, not at the edge) is left dangling. Since the prompt asks for Traditional Chinese titles and only soft-instructs "do not include quotes" with no enforcement, an LLM embedding a quoted rumor name mid-title (a natural Chinese headline style) produces a mangled, unbalanced title that gets shown to the user.

Suggested fix direction: I think maybe we don't need to forcefully strip these quote / brackets in code. If LLM decides to ignore our prompt and still generate quotes or brackets, maybe they have a reason for that.

adk/cofacts_ai/session_title.py
```diff
@@ -0,0 +1,120 @@
...
+    return None
+
+
+def _count_user_events(events) -> int:
+    return sum(1 for event in events if event.author == "user")
+
+
+def _last_writer_text(events) -> str:
+    for event in reversed(events):
+        if event.author != "writer":
```
MrOrz, line 59: nit: `_last_writer_text` hardcodes the literal author string "writer", a second independent coupling point to `ai_writer`'s `name=` field in `agent.py` with no shared constant.

`agent.py` already shows naming drift between the `ai_writer` variable/comments and the actual `name="writer"` field. If a future rename changes the `name=` field without updating this literal, `_last_writer_text` silently returns `""` forever; since the guard above only bails when both `user_text` and `result_text` are empty, title generation continues using only the user's message, degrading quality with zero exception, log, or test failure (the test file hardcodes "writer" too, so it wouldn't catch a real rename).

MrOrz: I think we can extract all agent names in a follow-up issue. We don't need to refactor in this pull request

--> #105

adk/cofacts_ai/session_title.py
MrOrz: Suggested fix direction: I think we can block user from editing session names when `isStreaming`. This would also help reducing ADK event / state changes while LLM is processing.

SametAtas:
Thanks, both catches were right.

1. Quote normalization: quotes are now stripped only as a matched pair, so a wrapper the
model adds is removed while quotes belonging to the title survive. Previously
他說「這是假的」 came back as 他說「這是假的. Added NormalizeTitleTest covering the wrapped
ASCII and CJK cases plus the unmatched-edge cases that were mangled (7/7 passing).

2. Race condition: agreed, the callback's write could land after a sidebar rename and
clobber it under last-write-wins. SessionItem now observes its own session's chat cache
(read-only, enabled: false, so the sidebar never fetches) and hides the rename affordance
while that session is streaming. This follows the existing precedent in
session.$sessionId.tsx, which already suppresses ADK state updates during streaming. I
verified the observer both reacts to a stream starting while mounted and reads correctly
when a session is already streaming at mount.

Verified: tsc --noEmit clean, and the 71 existing vitest tests still pass.

Two notes: I did not add a component test for the gate since the repo has no
component-test precedent yet, happy to add one if you would like. And `pnpm lint` fails
on a clean checkout too (eslint is not installed), which looks like the same thing #94
tracks.

MrOrz (approved):
LGTM! Thanks for the contribution!

PR #19, opened 2026-03-15
Enhanced Langfuse Integration: Trace Text Extraction & User Feedback

MrOrz:
This submission enhances the existing Langfuse integration to improve observability and user engagement.

Key changes:
1.  **Trace Readability (Backend):** Added a custom `TextExtractionSpanProcessor` in the Python ADK. It intercepts OpenTelemetry spans and extracts plain text from the complex JSON structures typically sent as input/output, making the Langfuse session view much easier to browse.
2.  **Trace ID Synchronization:** The backend now captures the active `trace_id` and includes it in the `custom_metadata` of its responses. The frontend extracts this ID and stores it within each chat message, creating a reliable link between the UI and backend traces.
3.  **User Feedback (Frontend):** Integrated the Langfuse Web SDK and updated the `AgentMessage` component. Users can now upvote or downvote agent responses, and these ratings are sent to Langfuse as `user_feedback` scores (1 or -1) associated with the specific trace ID of that message.
4.  **Dependency Management:** Added the `langfuse` JS SDK to the frontend dependencies.

These changes provide a comprehensive feedback loop and significantly better debugging capabilities for the Cofacts AI system.

Fixes #14

---
*PR created automatically by Jules for task [15597546857061770299](https://jules.google.com/task/15597546857061770299) started by @MrOrz*

Issue #14, opened 2026-03-12
Langfuse integration


MrOrz:
On top of current Langfuse integration, add:

- Extract text and image for trace input & output for easier browsing in session view
- Upvote / Downvote

MrOrz:
2, 3 --> #26 and https://github.com/cofacts/ai/pull/24

1 --> Leave it to @godgunman

PR #110, opened 2026-07-17
feat(instrumentation): extract plain text for Langfuse trace input/output

SametAtas:
Remaining backend half of #14: readable trace input/output. The other two parts
(trace-id sync via LangfuseTracingPlugin, feedback scoring via user-thumbs) are
already on master.

GoogleADKInstrumentor stamps input.value with the serialized run_async arguments
and output.value with the final Event JSON, so the Langfuse session view shows
raw JSON blobs. This adds TextExtractionSpanProcessor in instrumentation.py: at
span end it parses those payloads and replaces them with the joined text of
new_message.parts / content.parts, and flips the mime type to text/plain.
Thought parts are skipped (the writer runs with include_thoughts=True). Non-JSON
and no-text payloads (tool calls, image-only messages) are left untouched. Image
extraction is out of scope per the issue.

Kept backend-only, following the direction in #19 to isolate FE and BE changes.

Two deviations from the #19 sketch, both found by testing against the installed
packages:

1. on_end receives a ReadableSpan snapshot that has no set_attribute (OTel SDK
   1.38), so that approach raises AttributeError. The processor updates the
   span's attribute mapping in place, replacing existing keys only, wrapped in
   catch-and-log so a failure can never break span end.

2. Langfuse's BatchSpanProcessor is registered inside get_client(), before
   setup_instrumentation can add the rewrite processor, and processors run in
   registration order. A batch flush could therefore race the rewrite and
   export raw JSON. After registration the rewrite processor is moved to the
   front of the provider's processor list; if SDK internals ever change, this
   degrades to the previous order instead of failing setup.

Verification (offline, deterministic): 13 new tests in
tests/test_instrumentation.py replicate the production registration order and
export through an exporter that snapshots attributes at export() time, the way
OTLP serialization does. Covered: input and output rewrite, mime flip,
thought-part skipping, multi-part joining, rewrite-before-export ordering, and
negative controls (non-JSON, no-parts JSON, image-only, no input/output,
read-only attribute mapping, unexpected provider internals, and raw JSON being
exported when the reorder is absent). uv run pytest: 38 passed. ruff format,
ruff check, and ty: clean.

Pending: visual check of a real trace in the Langfuse session view.

### DIFF
--- a/adk/cofacts_ai/instrumentation.py
+++ b/adk/cofacts_ai/instrumentation.py
@@ -149,19 +149,22 @@ def _move_processor_first(provider: SDKTracerProvider, processor: SpanProcessor)
     public ordering API; on any surprise we leave the order unchanged, which
     degrades to that occasional race, not to an error.
     """
-    multi = getattr(provider, "_active_span_processor", None)
-    processors = getattr(multi, "_span_processors", None)
-    if (
-        multi is None
-        or not isinstance(processors, tuple)
-        or processor not in processors
-    ):
-        logger.warning("Could not reorder span processors; rewrite runs last.")
-        return
-    multi._span_processors = (
-        processor,
-        *(p for p in processors if p is not processor),
-    )
+    try:
+        multi = getattr(provider, "_active_span_processor", None)
+        processors = getattr(multi, "_span_processors", None)
+        if (
+            multi is None
+            or not isinstance(processors, tuple)
+            or processor not in processors
+        ):
+            logger.warning("Could not reorder span processors; rewrite runs last.")
+            return
+        multi._span_processors = (
+            processor,
+            *(p for p in processors if p is not processor),
+        )
+    except Exception:
+        logger.exception("Could not reorder span processors; rewrite runs last.")
 
 
 def setup_instrumentation():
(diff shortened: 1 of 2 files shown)


SametAtas:
Adopted the try/except around the attribute rewrite in 3ac1dfc. `_attributes`
is private SDK state, and OTel 1.38's synchronous span processor does not
isolate exceptions raised by processor `on_end` methods, so a mutation failure
could propagate out of every span end. The new test passes a read-only mapping
through `on_end` and verifies that no exception escapes and the original value
remains unchanged.

On the tracer-provider `isinstance` check, I left the existing pattern
unchanged. The cast-and-register idiom already exists on master for
`RootSessionSpanProcessor`. In the current Langfuse 4.0.0 path, `get_client()`
initializes or reuses the tracer provider and registers Langfuse's own span
processor before `auth_check()` runs; when the global provider is still a
`ProxyTracerProvider`, Langfuse installs an SDK `TracerProvider`. I kept that
existing assumption in this focused PR. Broader provider hardening can be
handled separately if preferred.

MrOrz:
cc/ @godgunman please check if it would affect your automated tool

Issue #84, opened 2026-06-06
Long multimodal runs hit ~300s request timeout with no final answer


MrOrz:
## Summary

When the writer↔verifier↔investigator loop runs long on a hard-to-adjudicate VIDEO/AUDIO article, the whole invocation gets cut off at **~300 s** (an infrastructure request timeout) and returns **no final answer** — `trace.output` is `null`, with no `ERROR`-level observation. The run is killed mid-loop, not by `max_llm_calls` (set to 500, nowhere near hit).

## Evidence

[Langfuse session `03826a6b`](https://langfuse.cofacts.tw/project/cmm0emerr0001qi07eugd0760/sessions/03826a6b-bbfd-4ea1-8e20-0a3f1bc31b98) — a Jingdezhen-ceramics craft video.

- Media delivery worked fine: 3 `verifier` passes, all with the correct `gs://` attachmentUrl.
- The video had no clear-cut false claim, so the writer kept re-querying `investigator` / `verifier` to decide whether it even counts as misinformation.
- Timeline: started `10:24:15.819`, last observation ended `10:29:14.586` → **298.77 s elapsed**, then the run stops. The writer's final action was a 4th `investigator` call that returned at `10:29:14`; no subsequent writer turn produced a final answer.

This is the latency/cost ceiling of the multimodal verification loop, surfaced while reviewing #72. It is **not** caused by the media-delivery changes in that PR (those worked correctly here).

## Things to investigate

1. **Where the ~300s limit lives** — Cloud Run request timeout, SSE/HTTP timeout, or another layer? Confirm the source and whether it can be raised.
2. **Convergence** — should the writer have a verification-round budget, or guidance to make a call once the evidence is sufficient rather than chasing certainty on ambiguous "is this even misinfo?" cases? (The two sibling sessions, [`42543ea6`](https://langfuse.cofacts.tw/project/cmm0emerr0001qi07eugd0760/sessions/42543ea6-a707-4843-90d7-2cf4bf6730a9) and [`cc9ed3bd`](https://langfuse.cofacts.tw/project/cmm0emerr0001qi07eugd0760/sessions/cc9ed3bd-65aa-405f-b799-d650880118e7), converged quickly because they had a clear adjudication point.)
3. **Graceful degradation** — if the timeout is unavoidable, can we emit a partial/"still working" result instead of dropping the answer entirely?

Refs: #72

SametAtas:
I traced the ~300 s cutoff from question 1.

[...]

PR #113, opened 2026-07-19
fix(deploy): set Cloud Run request timeout to 900s

SametAtas:
Part of #84 (question 1).

The revision template in service.template.yaml sets no
spec.template.spec.timeoutSeconds, so the Cloud Run service (preview
environments, per discussion in #84) uses the default 300 s request timeout.
Long multimodal runs are cut off mid-loop with no final answer.

This sets timeoutSeconds: 900. Verified offline: the rendered template
(envsubst with placeholder values) parses, the rendered request timeout is
900, and both startupProbe timeoutSeconds values remain 5. The live revision
value was not inspected; gcloud run services describe can confirm it after
deploy.

The agent-level budget work (items 2 and 3 in #84) is intentionally not part
of this PR.

### DIFF
--- a/service.template.yaml
+++ b/service.template.yaml
@@ -16,6 +16,7 @@ spec:
         autoscaling.knative.dev/maxScale: "3"
     spec:
       serviceAccountName: ${SERVICE_ACCOUNT_EMAIL}
+      timeoutSeconds: 900
       containers:
       - name: ingress
         image: ${FRONTEND_IMAGE}


MrOrz:
Seems duplicate with #112

SametAtas:
Confirmed. I missed #112 while preparing this. #113 is closed as a duplicate. Thanks.

Issue #105, opened 2026-07-09
refactor: Extract agent names into a constant


MrOrz:
Since name of the agents are used in conditions and prompts, we should extract them as constants so that we can keep track of them, and change them together when necessary.

_Originally posted by @MrOrz in https://github.com/cofacts/ai/pull/102#discussion_r3548454712_

PR #114, opened 2026-07-19
test: move session title tests into the pytest suite

SametAtas:
Follow-up to #102 and the CI introduced in #108.

pytest's `testpaths` is `cofacts_ai/tests`, so the session title tests from #102
(`cofacts_ai/session_title_test.py`) pass when run explicitly but are never
collected by CI: `uv run pytest` collects 25 tests on master, none from
session_title. A regression in the title callback would go unnoticed.

This moves them to `cofacts_ai/tests/test_session_title.py` and converts them
from unittest style to the suite's conventions (plain classes, bare asserts,
typed `make_*` fakes, module docstring). The `AI_WRITER_NAME` constant usage
from #105 is carried into the moved fixtures. No source files are touched.

Verified: `uv run pytest` now collects and passes 32 (25 + 7);
`ruff format --check`, `ruff check`, and `ty check` all clean.


MrOrz:
@claude /code-review

MrOrz (approved):

PR #118, opened 2026-07-22
Pre-fetch verifier URLs via url-resolver to stop source hallucination

MrOrz:
## Summary

Langfuse feedback analysis (130 `user-thumbs` items, 2026-04 → 2026-07) shows source hallucination — the verifier claiming a URL supports a claim when the link is dead or the page doesn't contain the info — is the single biggest failure mode in the pipeline (~21% of all feedback, ~34% of downvotes). Root cause: the verifier only ever reads pages through Gemini's black-box `url_context` tool, with no deterministic proof a page was actually fetched.

This PR adds a `before_model_callback` on `ai_verifier` that pre-fetches every plain web URL through [`cofacts/url-resolver`](https://github.com/cofacts/url-resolver) (a gRPC service using puppeteer + Readability.js) and injects the real cleaned body text into the verifier's context — text it cannot skip reading. A URL the resolver can't resolve at all (DNS failure / malformed) gets an advisory `[LINK NOT FOUND]` note rather than a hard ban, since `url_context` may still succeed where the resolver's simpler fetch failed.

- **`url_context` is kept**, not replaced — it's the only source for YouTube video content and page metadata (upload date, title). A `before_model_callback` isn't a tool, so this sidesteps the existing ADK constraint that forbids mixing built-in tools with function-calling tools on one agent.
- **Full page text is cached as an ADK artifact**, keyed by a hash of the URL. This doubles as the fetch cache (skips re-resolving across the verifier's ~2 model calls per turn, and across turns in the same session) and as a UI-visible record of exactly what the system read. `html` (the raw page, 100KB+) is dropped before it ever leaves the client — never sent to an LLM.
- **Injected text is budgeted, not truncated by default**: all resolved pages are injected in full up to a total char budget (`URL_RESOLVER_TOTAL_CHAR_BUDGET`, default 200k chars across the whole turn); only when a turn's total exceeds that does a max-min fair allocation trim the longest pages, leaving short ones untouched.
- **The source wrapper (`append_verifier_sources`) now builds `sources` as the union** of pages url-resolver resolved and `url_context`'s grounding chunks, deduped by URL. A genuinely dead URL is fetched by neither path, so it's silently absent from `sources` — no hard-coded ban list needed, and it can't reach the writer's `draft_factcheck_response` claim-sources gate.

## Changes

- `cofacts_ai/url_resolver/` — new package: vendored `.proto` files from `cofacts/url-resolver`, generated gRPC stubs (committed, so the runtime image only needs `grpcio`, not `grpcio-tools`), and an async client (`resolve_urls`) that buckets resolver errors into `DEAD` (the URL itself is bad) vs. `RESOLVER_CANT_FETCH` (resolver limitation — PDF, blocked, etc. — `url_context` may still succeed) vs. `TIMEOUT` / `RESOLVER_UNAVAILABLE` (no signal, never treated as dead).
- `cofacts_ai/agent.py` — new `inject_resolved_url_content` before-model callback on the verifier; `append_url_context_sources` rewritten to `append_verifier_sources`; verifier instruction updated to explain the pre-fetched `[RESOLVED PAGE]` / `[LINK NOT FOUND]` markers and the division of labour with `url_context`.
- `scripts/gen_protos.sh` — regenerates the vendored gRPC stubs from the `.proto` files.
- `pyproject.toml` — adds `grpcio` (runtime) and `grpcio-tools` (dev only, for codegen).
- `cofacts_ai/.env.example` — documents the four new `URL_RESOLVER_*` env vars.

## Deployment note

url-resolver is insecure gRPC with no auth and must stay internal-only. In production it's expected to share a docker-compose network with the ADK service (reachable by service name, default `URL_RESOLVER_ADDRESS=url-resolver:4000`); on Cloud Run it would run as an ADK sidecar (`localhost:4000`).

## Test plan

- [x] `uv run pytest` — 65 tests pass, including 3 new files covering the gRPC client (join-by-URL, error bucketing, transport failure handling), the injection callback (resolved/dead/unfetchable/excluded URLs, idempotency, artifact caching, budget truncation), and the source wrapper (dead-URL exclusion, union-with-grounding, dedup)
- [x] `uv run ruff check .` / `uv run ruff format --check .` — clean
- [x] `uv run ty check` — clean
- [ ] Manual end-to-end: run `cofacts/url-resolver` locally via Docker, point `URL_RESOLVER_ADDRESS` at it, and confirm a known-dead URL is reported unreachable (not cited) while a known-good URL's report quotes the injected page text

https://claude.ai/code/session_01QbfiuqaNfZEiXsEo6jGXNv

---
_Generated by [Claude Code](https://claude.ai/code/session_01QbfiuqaNfZEiXsEo6jGXNv)_


adk/cofacts_ai/agent.py
```diff
@@ -269,6 +302,243 @@ def inject_youtube_filedata(
...
+        titles: dict[str, str] = {}
+        canonicals: dict[str, Optional[str]] = {}
+
+        to_fetch: list[str] = []
+        for url in urls:
+            filename = _resolved_artifact_filename(url)
+            cached = await callback_context.load_artifact(filename)
+            if cached is not None and cached.inline_data and cached.inline_data.data:
+                text = cached.inline_data.data.decode("utf-8")
+                version_info = await callback_context.get_artifact_version(filename)
```
MrOrz: **Blocking:** `get_artifact_version` is `raise NotImplementedError` in ADK 1.26.0's `ForwardingArtifactService` — which is exactly the artifact service an `AgentTool`-hosted agent gets, i.e. `ai_verifier` in production. So the first artifact cache *hit* raises, the broad `except Exception` at the bottom of this callback swallows it, and the entire injection is abandoned: no `[RESOLVED PAGE]` parts at all, silently degrading the verifier back to `url_context`-only — the exact behaviour this PR exists to remove.

Timing makes it worse than a corner case: the cache is populated on the 1st model call of a verifier turn, so this fires on the 2nd model call of that same turn and on every verifier call afterwards in the session. Meanwhile `temp:cofacts_resolved_meta` from the earlier call can still be in state, so `append_verifier_sources` keeps reporting those pages as `sources` for a response that never saw their text.

The suite misses it because `FakeArtifactStore` implements `get_artifact_version`; the real plumbing does not.

Suggest not depending on `get_artifact_version` here — e.g. carry title/canonical inside the cached artifact body (a small header line or a JSON envelope), or wrap the call in a targeted `try/except NotImplementedError` that falls back to `titles[url] = url; canonicals[url] = None` rather than losing the page text.

---

adk/cofacts_ai/agent.py
```diff
@@ -269,6 +302,243 @@ def inject_youtube_filedata(
...
+                "title": titles[url],
+                "canonical": canonicals[url],
+            }
+
+        if not injected_parts:
+            return None
+        target_content.parts = list(target_content.parts) + injected_parts
+        if resolved_meta:
+            callback_context.state[RESOLVED_META_STATE_KEY] = resolved_meta
+    except Exception:
```
MrOrz: This blanket `except Exception` is what turns the `get_artifact_version` bug above into an invisible one, and it will do the same for a missing artifact service or a non-numeric `URL_RESOLVER_TOTAL_CHAR_BUDGET`: the failure mode is a silent, total revert to `url_context`-only, which looks identical to "no URLs in the message". Note the asymmetry with #123, where the deploy workflow deliberately hard-fails on a half-configured resolver to prevent exactly this kind of quiet degradation — the runtime path should be at least as loud.

Two suggestions:
- Narrow the guard to the parts that genuinely may fail against a live network (`resolve_urls`, artifact I/O) so a programming error surfaces instead of being logged and skipped.
- Even keeping the broad catch, clear `callback_context.state[RESOLVED_META_STATE_KEY]` on the failure path — otherwise stale meta from a previous call still feeds `sources`.

---

adk/cofacts_ai/agent.py
```diff
@@ -269,6 +302,243 @@ def inject_youtube_filedata(
...
+            continue
+        target_content = content
+        for url in found_here:
+            if url not in seen:
+                seen.add(url)
+                urls.append(url)
+    return urls, target_content
+
+
+def _url_already_injected(llm_request: LlmRequest, url: str) -> bool:
```
MrOrz: `_url_already_injected` can never return `True`, so it is dead code and the dedup the docstring promises doesn't happen.

The premise — "the request is rebuilt from conversation history each time" — is right, but the consequence is the opposite: ADK deep-copies event content when rebuilding `llm_request.contents` (`flows/llm_flows/contents.py:522`), so the parts this callback appends to `target_content` are mutations of the *request* object only. They are never persisted as an event, so nothing carries them into the next model call's contents.

Practical effect: on every extra model call in a turn, all the pages are injected again — up to `URL_RESOLVER_TOTAL_CHAR_BUDGET` (200k chars) of page text re-sent per call. Artifact caching keeps that from re-fetching, but not from re-sending.

The idempotency test passes only because it re-uses the same mutated `LlmRequest` object across both invocations, which production never does.

Either drop the helper and accept re-injection (it is functionally correct, just costly), or dedup off something that does survive — e.g. record injected URLs in `temp:` state alongside `RESOLVED_META_STATE_KEY` — and add a test that builds a fresh `LlmRequest` for the second call.

---

adk/cofacts_ai/agent.py
```diff
@@ -153,30 +158,58 @@ async def _resolve(chunk) -> Optional[str]:
     return _set_text_content(llm_response, serialized)
 
 
-async def append_url_context_sources(
+RESOLVED_META_STATE_KEY = "temp:cofacts_resolved_meta"
```
MrOrz: `temp:` state is not scoped per `AgentTool` call — ADK forwards the child's state delta up to the parent invocation, and the parent's state is what seeds the next child. So a later `ai_verifier` call that resolves nothing (all URLs `TIMEOUT`, or the injection aborted early per the `get_artifact_version` comment below) still sees the *previous* call's `RESOLVED_META_STATE_KEY` and emits those URLs as its own `sources`.

That is the failure this PR is meant to close: a claim citing pages the model never read in that call. Worth having `inject_resolved_url_content` write the key unconditionally — `state[RESOLVED_META_STATE_KEY] = resolved_meta` including the empty case, and on the `except` path — so each call's `sources` reflect only what that call resolved.

---

adk/cofacts_ai/agent.py
```diff
@@ -153,30 +158,58 @@ async def _resolve(chunk) -> Optional[str]:
...
         return None
 
+    resolved_meta = callback_context.state.get(RESOLVED_META_STATE_KEY) or {}
     sources_list = [
-        {
-            "title": (chunk.web and chunk.web.title) or "Unknown Source",
-            "url": chunk.web.uri if chunk.web else None,
-        }
-        for chunk in chunks
+        {"title": meta.get("title") or url, "url": meta.get("canonical") or url}
```
MrOrz: Minor, same union: resolver sources are keyed by `meta.get("canonical") or url` while the `url_context` branch below keys `seen_urls` by the grounding chunk's URI (the requested URL). When a page redirects or declares a different canonical, the same page lands in `sources` twice under two URLs. Deduping on a normalized form of both, or keeping the requested URL as the identity and the canonical as a display detail, avoids it.

---

adk/cofacts_ai/agent.py
```diff
@@ -269,6 +302,243 @@ def inject_youtube_filedata(
...
+    target_content = None
+    for content in llm_request.contents:
+        if content.role != "user" or not content.parts:
+            continue
+        found_here = []
+        for part in content.parts:
+            if not part.text:
+                continue
+            for match in _HTTP_URL_RE.findall(part.text):
+                url = match.rstrip(_URL_TRAILING_PUNCT)
```
MrOrz: `rstrip(_URL_TRAILING_PUNCT)` strips a trailing `)` that is part of the URL — Wikipedia disambiguation links are the common case (`.../wiki/Mercury_(planet)`). The truncated URL 404s, url-resolver reports it `DEAD`, and the verifier is handed a `[LINK NOT FOUND] … do NOT claim this link supports any claim` note about a perfectly good source. Because `rstrip` strips a *set*, `…_(planet).` loses both characters too.

The usual fix is to only strip a trailing `)` when the URL has no matching `(`, i.e. balance the parens before stripping, and leave the other punctuation handling as-is.

---

adk/cofacts_ai/agent.py
```diff
@@ -269,6 +302,243 @@ def inject_youtube_filedata(
...
+                    canonicals[r.url] = r.canonical
+                    await callback_context.save_artifact(
+                        filename=_resolved_artifact_filename(r.url),
+                        artifact=genai_types.Part(
+                            inline_data=genai_types.Blob(
+                                mime_type="text/plain",
+                                data=r.summary.encode("utf-8"),
+                            )
+                        ),
+                        custom_metadata={"title": r.title, "canonical": r.canonical},
```
MrOrz: `r.title` and `r.canonical` are `Optional[str]`, and `GcsArtifactService` stringifies `custom_metadata` values when it writes them as blob metadata. A page with no title/canonical therefore round-trips as the literal string `"None"`, and on the next cache hit `titles[url]`/`canonicals[url]` become `"None"` — which reaches the UI as `{"title": "None", "url": "None"}` in `sources` (`or url` doesn't help; `"None"` is truthy).

Dropping the `None` entries before saving, and treating `"None"` defensively on read, keeps the fallbacks working.

---

adk/cofacts_ai/url_resolver/client.py
```diff
@@ -0,0 +1,222 @@
...
+from typing import Optional
+
+import grpc
+
+from ._generated import resolve_error_pb2, url_resolver_pb2, url_resolver_pb2_grpc
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_ADDRESS = "url-resolver:4000"
+DEFAULT_TIMEOUT = 30.0
```
MrOrz, line 28: 30s is a deadline for the *whole* streaming call, not per URL — it is passed straight to `stub.ResolveUrl(...)`, which covers every URL in the batch. With `DEFAULT_MAX_URLS = 20` and url-resolver's `SCRAPE_MAX_CONCURRENCY` default of 3, that is ~7 sequential rounds; against a remote Cloudflare browser (per #123, now the pinned backend) each round is a session acquire + navigate + Readability pass. A message with more than a handful of links will realistically hit the deadline, and everything not yet returned falls into `TIMEOUT` → nothing injected → `url_context`-only, silently.

`URL_RESOLVER_TIMEOUT` exists but `service.template.yaml` in #123 doesn't set it, so production runs on this default. Worth either raising the default substantially (the Cloud Run request timeout is 900s) or making it scale with `len(deduped_urls)`, and setting it explicitly in the deploy template.

---

adk/cofacts_ai/url_resolver/client.py
```diff
@@ -0,0 +1,222 @@
...
+    )
+    timeout = timeout or float(os.environ.get("URL_RESOLVER_TIMEOUT", DEFAULT_TIMEOUT))
+
+    deduped_urls: list[str] = []
+    seen = set()
+    for url in urls:
+        if url not in seen:
+            seen.add(url)
+            deduped_urls.append(url)
+    deduped_urls = deduped_urls[:max_urls]
```
MrOrz: URLs past `max_urls` are dropped here and get no `ResolvedUrl` at all, so the caller cannot tell "never attempted" from "attempted and unresolvable". In `inject_resolved_url_content` they simply produce no injected part — indistinguishable from `RESOLVER_CANT_FETCH`/`TIMEOUT` — and the verifier is left to assume `url_context` covers them, with no note saying nothing was pre-fetched.

Returning a distinct status (or at least a `NOT_ATTEMPTED` bucket) for the truncated tail would let the callback inject an explicit "not pre-fetched, rely on url_context" note, consistent with how the other non-fatal buckets are surfaced.

---

MrOrz (commented):
Review of the verifier URL pre-fetch. 110 tests pass, so everything below is outside what the suite covers — in two cases because the fakes are more capable than the real ADK plumbing.

Two are worth fixing before merge, since they undo the PR's core guarantee in production:

- `agent.py:453` — `get_artifact_version` is `NotImplementedError` under the `ForwardingArtifactService` an `AgentTool`-hosted agent actually gets, so every artifact cache hit aborts the whole injection through the broad `except` and reverts the verifier to `url_context`-only.
- `agent.py:161` — `temp:cofacts_resolved_meta` survives across `AgentTool` calls, so a call that resolves nothing can still emit the previous call's URLs as its `sources`.

The rest are smaller: the broad `except Exception` that hides the above, `_url_already_injected` being unreachable dead code (200k chars re-injected per extra model call), `)`-truncated Wikipedia URLs flagged as dead links, `None` metadata round-tripping as the string `"None"`, a canonical-vs-requested dedup gap in `sources`, and two url-resolver client defaults (a 30s whole-stream deadline for up to 20 URLs, and silently dropped URLs past `max_urls`).

Checked and cleared: `reply.url` echo, proto enum bucketing, `_water_fill` allocation math, per-`AgentTool` isolation of `sources` accumulation, and the always-wrap change in `append_verifier_sources`.

---

adk/cofacts_ai/agent.py
MrOrz: Fixed in `8181725`. Confirmed the premise first: `get_artifact_version` support on `ForwardingArtifactService` landed in [adk-python 2.5.0](https://github.com/google/adk-python/releases/tag/v2.5.0), and `adk/uv.lock` pins 1.26.0 — so it raises here, exactly as described.

Took the first of your two suggestions (metadata inside the artifact body) rather than the `try/except NotImplementedError` fallback, because that fallback keeps the feature alive but drops every cached page's title, which surfaces in the UI as `sources` entries labelled with a bare URL. The body is now a JSON header line plus the page text, via `_encode_resolved_artifact` / `_decode_resolved_artifact`, and the version API is not called at all.

Two things that came with it:

- It also removes the `"None"` round-trip you raised separately on line 478 — `GcsArtifactService` stringifies `custom_metadata`, so an untitled page came back as the truthy string `"None"` and defeated the `or url` fallback. JSON preserves a real `null`. That thread is yours to close; I have not touched it beyond this.
- Artifacts written before the envelope are raw text and preview deploys have already written some, so `_decode_resolved_artifact` treats a body whose first line is not a JSON object as page text with no metadata. Raising there would have reintroduced the same silent-abandonment failure.

On the suite missing it: `FakeArtifactStore.get_artifact_version` now **raises**, matching the real service, so every test in the file exercises the production plumbing rather than a fake that is kinder than reality. Four tests fail without the fix — including the pre-existing `test_second_call_reuses_artifact_cache_not_network`, which is the one that should have caught this originally. Full suite is 113 passed, ruff and `ty` clean.

---

adk/cofacts_ai/agent.py
MrOrz: Fixed in `b5d8bf9`. Confirmed the mechanism in ADK first: `agent_tool.py:236-244` seeds the child session from the parent's `tool_context.state.to_dict()`, and `:254-256` forwards the child's delta back up — so `temp:` is not scoped per call, exactly as you said.

Three paths left the key untouched, not one: the `if not urls ...` early return, the `if not injected_parts` early return, and the blanket `except`. So the key is now cleared **on entry**, before the `try`, and assigned unconditionally at the end. Your suggested one-line change at the write site would have covered the third case only.

Worth stating explicitly because it is the obvious objection: **this costs no extra fetches.** The fetch cache is the artifact (`resolved-<sha1>.txt`) and the fetch decision only calls `load_artifact`; this key is purely the before→after handoff, since `append_verifier_sources` cannot see the injected request parts. There is a test asserting the second call still serves from the artifact cache (`resolve_urls` awaited once) so that stays true.

Two existing tests asserted the key was *absent* after a no-op call. That assertion encoded the leak, so they now assert present-and-empty.

Worst case this closes, for the record: `ai_verifier` runs once per claim, so a claim with no links was citing the previous claim's URLs.

---

adk/cofacts_ai/agent.py
MrOrz: Fixed in `b5d8bf9`, and the investigation turned up something bigger than the paren case.

**rumors-site already solved this.** `lib/text.tsx` extracts URLs with `linkifyjs.tokenize()`, and `lib/__tests__/text.tsx` pins the behaviour under `parses half-width brackets correctly` / `parses full-width brackets correctly`. Its snapshots handle balanced parens, wrapping parens, an unbalanced trailing `)`, **and the full-width `（）` forms**. Both of those exact strings are now tests here, so the two codebases agree on what counts as a URL.

**The wider bug:** `_HTTP_URL_RE`'s `[^\s\"'<>]+` treats full-width punctuation as URL characters, and Chinese has no spaces — so `https://example.com/a，另一個` was extracted as a single URL swallowing the following words. It resolves `DEAD` and the verifier gets your `do NOT claim this link supports any claim` note about a perfectly good source. Given Cofacts' messages are mostly Chinese, that fires far more often than the Wikipedia case.

That also means the fix you suggested (balance the parens before stripping) isn't sufficient on its own, and neither was my first instinct of adding full-width punctuation to `_URL_TRAILING_PUNCT` — a trailing strip can't repair a URL that has already run into the next words. So: CJK sentence punctuation is excluded from the match itself, and the `rstrip` is replaced by a bracket-balancing trim that keeps `.../Mercury_(planet)` while dropping the wrapper in `(https://example.com)`, full-width included.

I tried `linkify-it-py` before hand-rolling it, and rejected it on evidence — it gets half-width brackets right but keeps a trailing full-width `）` and swallows following words in CJK prose exactly as the old regex did. It would have fixed the case you named and none of the ones that dominate here.

`_URL_TRAILING_PUNCT` stays in `media_filedata.py` for its own use. Note it has the same latent problem there (`_COFACTS_MEDIA_URL_RE` also ends in `[^\s\"'<>]+`, so a trailing `。` sticks and `signed_url_to_gs` derives a wrong blob name) — much narrower, since those URLs are machine-generated and the regex is domain-anchored. Left alone as a separate follow-up rather than widening this thread.

---

adk/cofacts_ai/agent.py
MrOrz: Fixed, as a side effect of the `get_artifact_version` thread rather than on its own: `custom_metadata` is not written at all any more, so there is nothing left to stringify. `_encode_resolved_artifact` packs `{"title", "canonical"}` as a JSON header line inside the artifact body, which preserves a real `null` — so an untitled page decodes to `None` and the `or url` fallback fires as intended, instead of the truthy string `"None"` reaching the UI as a source title.

Verified on the current head: no `custom_metadata` write remains anywhere in `adk/cofacts_ai/` (only the two docstring mentions in `resolved_pages.py` explaining why it is avoided).

The code has since moved out of `agent.py` into `adk/cofacts_ai/resolved_pages.py` (`e54e208`), which is why this thread shows as outdated. Resolving.

---

adk/cofacts_ai/agent.py
MrOrz: Still fixed after the rebase onto `master`, just at new coordinates: the fix commit is now `bbf454e` (was `b5d8bf9` before the rebase, that sha is unreachable), and the code moved out of `agent.py` into `adk/cofacts_ai/resolved_pages.py` in `e54e208` — which is why this thread shows as outdated. The clear-on-entry write and its comment came across the move unchanged. Resolving.

---

adk/cofacts_ai/agent.py
MrOrz: Status after the rebase — **half of this is done, keeping the thread open for the other half.** The code now lives in `adk/cofacts_ai/resolved_pages.py` (`e54e208`), which is why the thread shows as outdated; the `except Exception` is at the end of `inject_resolved_url_content` there.

- Done: the stale-meta point. `RESOLVED_META_STATE_KEY` is cleared on entry, before the `try` (`bbf454e`), so the `except` path — and both early returns — leave the key describing only the current call.
- Still open: narrowing the guard. `_extract_web_urls`, `_water_fill` and the `int(os.environ.get("URL_RESOLVER_TOTAL_CHAR_BUDGET", ...))` parse are all still inside the `try`, so a bad env var or a programming error is logged and skipped rather than surfaced, and the result is indistinguishable from "no URLs in the message".

---

adk/cofacts_ai/agent.py
MrOrz: Status after the rebase — **unchanged, still applies verbatim.** Keeping this open; it is only marked outdated because the code moved out of `agent.py` into `adk/cofacts_ai/resolved_pages.py` (`e54e208`). The helper is now at `resolved_pages.py:233` and its caller at `:293`, both byte-identical to what this thread quotes, and the idempotency test still re-uses the same mutated `LlmRequest` across both invocations.

---

MrOrz: Fixed in `ab296ca` by taking your first option — the helper is gone, and with it the `already_injected` set, the meta-only cache-hit branch it needed, and the comment explaining why a cache hit implies a resolved marker. Net **−111/+88** in `resolved_pages.py`.

Confirmed the premise first: `flows/llm_flows/contents.py:522` is `content = copy.deepcopy(event.content)`, and the parts this callback appends are never written back as an event, so nothing carries them into the next call's contents. `_url_already_injected` could not return `True` in production.

**Your second option would have been actively wrong, and it's worth writing down why**, because it's the intuitive fix. Deduping off `temp:` state *would* survive the rebuild — but the request it would suppress injection for no longer contains the page text. The 2nd model call of a turn is the one that writes the final report, so skipping it there would leave the verifier reasoning about pages it can no longer see: url_context-only, silently, which is the failure this PR exists to remove. Re-injection isn't a cost we're tolerating for lack of a better option; it's the only correct behaviour once contents are rebuilt per call. That reasoning is now in the docstring so the helper doesn't get reinvented.

Both tests you flagged now build a **fresh `LlmRequest`** for the second call, and assert the invariant that actually matters — the page is re-injected, the *fetch* is not repeated (`resolve_mock.assert_awaited_once()`, served from the artifact cache). `test_second_model_call_keeps_meta_for_cached_urls` also still pins that the cache hit rebuilds `RESOLVED_META_STATE_KEY` with the cached title, so the call producing the final response doesn't drop resolver-fetched pages from `sources`.

Cost is unchanged from what you measured: up to `URL_RESOLVER_TOTAL_CHAR_BUDGET` re-sent per model call, no re-fetch. I've left it there rather than trading correctness for tokens.

I also corrected the decision record, which claimed "the callback is idempotent across a turn" — that was describing the behaviour of code that never ran.

---

adk/cofacts_ai/agent.py
MrOrz: Other half done in `ab296ca` — the guard is now narrowed as you asked.

`_extract_web_urls`, the `URL_RESOLVER_TOTAL_CHAR_BUDGET` parse, `_water_fill` and all part assembly are outside the `try`; only `load_artifact`/`save_artifact` and the `resolve_urls` call remain inside. The split is by whose fault the failure is: a resolver outage is the network's and must degrade to url_context rather than fail the turn, while a bad env var or a bug in extraction or budgeting is ours and now raises instead of masquerading as a message with no links. That matches #123's deploy-time hard-fail rather than being quieter than it.

The state-key invariant survives the restructure, which was the thing worth checking: `state[RESOLVED_META_STATE_KEY] = {}` still runs before any of this, so the new `return None` in the `except` *and* a now-possible propagating raise both leave the key describing only the current call. No path can inherit the previous call's meta.

`test_unparsable_char_budget_raises` pins it — `URL_RESOLVER_TOTAL_CHAR_BUDGET=lots` now raises `ValueError` out of the callback instead of being logged and skipped.

---

adk/cofacts_ai/agent.py
MrOrz: Fixed in `ab296ca`. `seen_urls` now claims **both** identities of a resolved page — the requested URL and the canonical — while `sources` still lists it under the canonical, which keeps your second suggestion's display behaviour without needing URL normalisation.

I went with registering both rather than normalising, because normalising would have to guess which differences are cosmetic (`utm_*`, trailing slash) and which are meaningful; here the resolver has already told us the two names refer to one page, so there's nothing to infer.

`test_dedups_when_canonical_differs_from_requested_url` covers it: a page requested as `https://a.com/?utm_source=line` with canonical `https://a.com/article`, grounded by url_context under the requested URL, yields exactly one source under the canonical. It fails on the previous code with two entries.

---

adk/cofacts_ai/url_resolver/client.py
MrOrz: **Not changed in this pass — deliberately leaving the call to you, since it's your deployment.** Flagging rather than deciding, because I don't think this is a code question.

The analysis holds and I'm not disputing it: `timeout` goes straight to `stub.ResolveUrl(...)` (`client.py:157`), so it is one deadline for the whole batch, and everything unreturned becomes `TIMEOUT` → nothing injected → url_context-only, silently. That is the failure mode this PR exists to remove, arriving through the back door.

What stops me flipping the default unilaterally:

- Production is already covered — #123 sets `URL_RESOLVER_TIMEOUT=120`, so the `30.0` only governs local dev and compose, where the resolver is a local browser rather than a remote Cloudflare one and the arithmetic is much friendlier.
- Picking the right number is a question about the deployed topology (which browser backend, expected link counts per message, how much of the Cloud Run 900s you want a single verifier call to be able to consume), not about this code. `120` to match #123 is the obvious answer, but "scale with `len(deduped_urls)`" is the better one and changes the shape of the knob.

Happy to do either immediately — say which and it's a one-line change plus `.env.example`. My own preference is a floor plus per-URL increment (e.g. `30 + 10 * len(deduped_urls)`, capped), because it keeps short messages fast while not silently truncating long ones, but that is a real behaviour change and I'd rather not make it on my own judgement while the numbers depend on infrastructure you're still tuning.

Separately worth noting: this thread's failure mode is now *slightly* louder than when you wrote it. `ab296ca` narrowed the guard so a misconfigured budget or a bug raises — but a timeout still degrades quietly, and correctly so, since a resolver hiccup must never be mistaken for proof a URL is dead.

---

adk/cofacts_ai/url_resolver/client.py
MrOrz: **Confirmed as a real gap, but not changed in this pass — it is the one open thread whose fix only adds surface area.** This round was scoped to review points that shrink the diff or fix a bug in what's already here; a new status plus a new injection branch is neither, so I'm leaving it visible rather than quietly growing the PR.

Traced it to be sure of the boundary. `deduped_urls = deduped_urls[:max_urls]` (`client.py:151`) is the *only* way a URL leaves `resolve_urls` with no `ResolvedUrl` at all — every URL that is actually attempted comes back, since the stream-join fills unreturned ones with `TIMEOUT`. So the ambiguity is exactly the over-cap tail and nothing else, which makes it narrower than "indistinguishable from `RESOLVER_CANT_FETCH`/`TIMEOUT`" suggests, though your conclusion is right: the verifier gets silence and assumes url_context has it covered.

When it's picked up, it needn't cost a new enum member. The callback already knows `to_fetch`, so the tail is derivable at the call site:

```python
not_attempted = [u for u in to_fetch if u not in {r.url for r in results}]
```

and those can take a note in the same family as the existing `[NOTE]` — worth its own wording, since "we didn't look" is a different claim from "we looked and couldn't fetch". That's roughly 8 lines and a test, entirely additive.

Impact is low enough that I'd rather it not gate this PR: `ai_verifier` runs once per claim, so hitting 20 distinct non-YouTube, non-Cofacts-media URLs in a single claim's request is rare. Happy to add it here if you'd prefer it landed together — otherwise it's a clean follow-up issue and I'll leave this thread open as the record.

---

MrOrz:
## Manual verification: the url-resolver + Cloudflare browser integration works, but only a minority of URLs actually reach it

Tested PR #123's preview deployment end-to-end (real chat → writer → verifier → url-resolver), which is directly relevant to the two open threads on this PR's `url_resolver/client.py`.

**Test 1 — real CNA news URL**: sent a claim citing a live CNA article. The verifier's `[RESOLVED PAGE]` content came back short (~200 chars) and turned out to be an exact match for the page's `<meta name="description">` tag. Traced it to `resolveUrls.js`'s gate:

```js
fetchResult = await parseMeta(unshortened);       // plain HTTP fetch (unfurl.js)
if (fetchResult.isIncomplete) {                   // canonical/title/summary/topImageUrl any undefined
  fetchResult.merge(await scrap(unshortened));     // only THEN does the headless browser run
}
```

CNA (and the 3 other real-world URLs tried: president.gov.tw, ndc.gov.tw, setn.com) all have `og:image` + `meta description`, so `isIncomplete` is `false` and `scrap()` — the Cloudflare Browser Rendering path — is **skipped entirely**.
Langfuse trace: https://langfuse.cofacts.tw/project/cmm0emerr0001qi07eugd0760/traces/d11d76db142bf9fb26019ef398eb4d16

**Test 2 — forced the browser path**: retried with a URL that structurally can't satisfy `parseMeta()` (a zh.wikipedia.org article — no `meta description`, no `og:description`, so `summary` is always `undefined` → `isIncomplete` is always `true`). This time the injected `[RESOLVED PAGE]` text was 16,545 chars of full article body (multiple sections, footnote markers like `[4][5][6]`) — only obtainable via `scrap()` actually launching Puppeteer against Cloudflare Browser Rendering and running Readability.js on the rendered DOM. The extracted content was accurate enough that the verifier caught a real discrepancy in the test claim (the article says 臺北縣萬里鄉, not 台北市萬里區).
Langfuse trace: https://langfuse.cofacts.tw/project/cmm0emerr0001qi07eugd0760/traces/c0914364409e974baf4eeb6fac680715

**Takeaway for the two open threads here**: the Cloudflare-backed browser path is real and does get exercised in production — just for a minority of URLs (anything with decent SEO meta tags skips it via `parseMeta`). That actually sharpens both concerns:
- The 30s whole-batch deadline only bites when several URLs in one request lack SEO meta (e.g. a message full of Wikipedia/forum/raw-doc links) and all fall through to `scrap()` — rarer than "every URL", but real, and currently invisible until it happens.
- URLs dropped past `max_urls` silently look identical to "resolver tried and got nothing," same as before.

Going to address both now.

---

adk/cofacts_ai/url_resolver/client.py
MrOrz: Fixed in 79b153b: `resolve_urls()` now scales the deadline with `len(to_send)` (`DEFAULT_PER_URL_TIMEOUT` = 10s/URL, floor the original 30s) unless `URL_RESOLVER_TIMEOUT` is set explicitly, so a batch that needs the browser fallback for several URLs no longer shares one flat 30s budget. Left uncapped rather than added to the deploy template — at the `URL_RESOLVER_MAX_URLS` default (20) that's 200s, well inside Cloud Run's 900s request timeout, and a config knob felt like the wrong lever for a bug in the default itself.

One correction while I was in there: I couldn't find `SCRAPE_MAX_CONCURRENCY` anywhere in the `cofacts/url-resolver` source — `resolveUrls.js` runs all requested URLs through `Promise.all` with no concurrency limiter I could find, so the "~7 sequential rounds" estimate doesn't hold, though the core concern (one shared deadline, real browser cost per URL) is still correct and is what this fixes.

Also manually confirmed the browser fallback this timeout budgets for is real: on #123's preview deployment, a real CNA URL resolved via url-resolver's `parseMeta()` fast path and never touched the browser (its `[RESOLVED PAGE]` text matched the page's `<meta name="description">` verbatim), while a Wikipedia URL — which structurally can't satisfy that fast path (no meta description) — did exercise the actual Puppeteer/Cloudflare `scrap()` path and returned 16.5k chars of real article body. Traces: [CNA/fast-path](https://langfuse.cofacts.tw/project/cmm0emerr0001qi07eugd0760/traces/d11d76db142bf9fb26019ef398eb4d16), [Wikipedia/browser-path](https://langfuse.cofacts.tw/project/cmm0emerr0001qi07eugd0760/traces/c0914364409e974baf4eeb6fac680715).

adk/cofacts_ai/url_resolver/client.py
MrOrz: Fixed in 79b153b: added `ResolveStatus.NOT_ATTEMPTED` for URLs dropped by `URL_RESOLVER_MAX_URLS` before ever being sent — they now carry an `error` explaining why. `inject_resolved_url_content` gives them their own branch (mirroring `RESOLVER_CANT_FETCH`): an archived Cofacts copy wins if one exists, otherwise an explicit `[NOT PRE-FETCHED] <url>` note is always injected — unlike `TIMEOUT`/`RESOLVER_UNAVAILABLE`, which stay silent on purpose since those causes are hiccups that could be mistaken for a dead link, whereas "capped by max_urls" is a certain, known cause. Also updated the verifier's system instructions to document the new marker, and added tests covering the cap+timeout interaction, the whole-call-failure+cap interaction, and the archived-copy-wins-over-the-note case.

MrOrz:
@claude code-review

PR #122, opened 2026-07-30
docs(adr): record why Langfuse under-reports Vertex AI cost by ~2x

MrOrz:
## Why

July 2026 Vertex AI spend on `ocf - ikalatv` was **$45.95**, but Langfuse reported only **$20.44** of Gemini cost across the two projects — 56% of real spend invisible. That makes Langfuse useless for budgeting and, worse, useless for noticing a runaway agent.

The instinct was lost traces. **It wasn't.** Pulling all 1,735 July `GENERATION` observations from the `cofacts.ai` project and re-pricing them at rates derived from the billing CSV accounts for **$45.11 of the $45.95 bill (98%)**. The tokens are already in Langfuse — they're being priced at $0. This is a cost-attribution bug, not data loss, which is why the fix belongs in how usage is mapped rather than in how spans are exported.

**No runtime code changes.** The ADR is `status: proposed` and records the recommended fix; this PR is the finding plus the tooling to re-check it.

## Root causes

| # | cause | share of gap |
|---|---|---|
| 1 | `toolUsePromptTokenCount` never priced | ~43% |
| 2 | 290 generations (16.7%) arrive with no model name → whole call $0 | ~29% |
| 3 | thinking tokens land under an unpriced key name | ~22% |
| 4 | `gemini-3.5-flash` never resolved to a definition (permanently $0) | ~7% |

**Cause 1** is the big one. Gemini reports `toolUsePromptTokenCount` *separately from* `promptTokenCount`, counts it inside `totalTokenCount`, and Google bills it at the normal text-input rate — but no OTel attribute carries it, so it only ever reaches Langfuse inside `total`, a key with no price. One observation carried 2,955,369 billed tokens and was priced as 2,016 (**$0.0028**). **77.6% of July generations have `total != input + output`.**

Grouping the unpriced tokens by `gen_ai.agent.name` localizes it precisely — it's `url_context`, not grounding in general:

| agent | generations | input priced | **tool-use (unpriced)** | **reasoning (unpriced)** |
| --- | --: | --: | --: | --: |
| `verifier` (`url_context`) | 175 | 4,200,299 | **29,279,079** | 952,633 |
| `investigator` (`google_search`) | 197 | 4,841,309 | 0 | 469,376 |
| `writer` | 510 | 15,887,776 | 0 | 229,390 |
| 4× `proofreader_*` | 459 | 351,161 | 0 | 578,643 |

**Cause 2** is localized to the **streamed root agent**. All 290 are children of an `agent_run [writer]` span; in an affected trace the sub-agents running concurrently are instrumented perfectly. openinference patches `base_llm_flow.trace_call_llm` and writes to `get_current_span()`, while ADK's own `trace_call_llm` writes to the `span` **passed to it** — and on the writer's spans every ADK attribute is missing from the whole trace, so the two are writing to different span objects. That also explains why it's the *model name* that vanishes: the surviving openinference attributes are exactly those set outside its `if llm_request:` branch. User interruption was considered and ruled out (24/25 affected traces have a final answer — more than clean traces).

**Cause 3** is a one-key mismatch: the instrumentor emits `completion_details.reasoning`, but the managed Gemini definitions price `output_reasoning` / `thoughtsTokenCount`, and Langfuse matches usage keys *exactly*. The prices were never wrong — only the key name.

Counteracting all of the above, Langfuse *over*-charges ~$3.7 on cached tokens (it prices them at full rate while GCP discounts 90%), which is why the naive gap reads $25.5 rather than $28.6.

Ruled out: sampling (none configured), lost traces (token capture reconciles at ~96%), and the two Flash Lite names in the dashboard (just `rumors-api` replacing a retired `-preview` id on Jul 17).

## Two things reviewers will want to know up front

**This is the officially recommended integration.** Langfuse's [Google ADK guide](https://langfuse.com/integrations/frameworks/google-adk) recommends exactly what `adk/instrumentation.py` does, and presents no alternative. Nothing here is a misconfiguration. What differs is the regime the guide was validated in — a notebook, one agent, one run, non-streaming, success measured by "spans appear in the UI" — versus a long-lived FastAPI service running a streamed multi-agent tree where the number that matters is a dollar figure. The guide never mentions token usage, cost, or streaming.

**A pricing table can't fix it,** which is the first thing anyone proposes, and a plausible-sounding claim circulates that `toolUsePromptTokenCount` is automatically aggregated into the input usage type. Three independent reasons it isn't: the value reaches Langfuse in **no attribute at all** (the span's full 24-attribute set doesn't contain it — it survives only as an arithmetic residue inside `total`); the OpenInference spec defines exactly eight `llm.token_count.*` attributes and **none is for tool use**, so it's a vocabulary gap needing a spec change before a library change; and `total` is a **reserved derived key** ("spans all buckets and equals their sum"), so it can't be priced as a workaround.

## What's here

- **`docs/decisions/20260730-langfuse-usage-mapping.md`** — MADR weighing five options. Chosen, in two complementary parts: **own the mapping** in the existing `LangfuseTracingPlugin` (reading `llm_response.usage_metadata` directly — the only thing that reaches causes 1, 2 and 5), **and file the reasoning-key gap upstream** with Langfuse, since that one is an allowlist omission on a name the spec already defines, with a merged precedent in [langfuse#13571](https://github.com/langfuse/langfuse/issues/13571) → #13572. Rejected: waiting on upstream for cause 1 (needs spec, library and Langfuse releases in sequence); custom model definitions for the main fix (kept only for pre-registering a model id before deploying it, as a safety net for when the instance's managed model list lags Google — exactly what made `gemini-3.5-flash` silently $0); and dropping openinference for ADK's native OTel export, which is **worse** — ADK sets only input and output token counts and no total, so cause 1 would become *undetectable* rather than merely unpriced.
- **`adk/scripts/langfuse_gcp_reconcile.py`** — reproduces the analysis and doubles as a regression check. Derives effective $/Mtok *from the billing CSV* rather than hardcoding list prices, so it survives repricing and new models.

## Verification

```
cd adk && uv run python scripts/langfuse_gcp_reconcile.py \
    --billing-csv <export grouped by SKU> --from 2026-07-01 --to 2026-08-01
```

Reproduces the numbers above: $45.947 billed / $39.880 reconstructed for this project / $17.305 reported, 1,346/1,735 bucket mismatch, 290 unpriced models. June reproduces the same shape (85.1% mismatch), so it isn't a one-month artifact.

Two health metrics need no CSV and work as a standing check — `input + output + input_cached_tokens + output_reasoning == total` should hold for every generation, and today fails on 77.6%:

```
uv run python scripts/langfuse_gcp_reconcile.py --from … --to … \
    --max-bucket-mismatch 0.02 --max-unpriced-models 0     # exits 1 on breach
```

`ruff check` / `ruff format --check` / `ty check` / repo-wide `prettier --check .` / `pytest` (32 passed) all pass.

Two caveats the script surfaces for itself, both easy to misread otherwise:

- A Langfuse API key is **project-scoped** while the bill covers the whole account, so it prints per-family coverage. Flash Lite shows 26% because `rumors-api` is a separate Langfuse project — not a mapping bug.
- The CSV's line items sum to $46.326 while its own "Filtered total" says $45.947, a rounding artifact in Google's export. The script reports the stated total and flags the difference.

Note Langfuse computes cost **at ingestion and never backfills**, so July stays wrong regardless — any fix is judged only on data ingested after deploy.

## Follow-ups recorded but not fixed here

- `rumors-api` `src/graphql/util.js:940-946` reads only `promptTokenCount`/`candidatesTokenCount`, dropping `thoughtsTokenCount` and the per-modality split — significant when Vertex bills Flash Lite audio input at 2× text. Its `transcribeAV` also never `.end()`s a generation when the API call throws.
- Confirm the streaming trigger behind cause 2 with a live repro. If confirmed it deserves an openinference issue in its own right — it silently un-instruments the root agent of every streamed ADK deployment, well beyond our cost gap.
- Flush on shutdown in both services (neither calls it today) plus `cpu-throttling: "false"`. Client aborts are also unguarded end to end — not the cause of anything found here, but left to chance.
- `adk/cofacts_ai/session_title.py` calls a raw `genai.Client()` outside the instrumentor.
- Open question: `gemini-3.5-flash` ran Jul 12–15 and cost $1.70, but that model id is nowhere in this repo on `master`. Worth finding which revision ran it.

### DIFF
--- a/adk/scripts/langfuse_gcp_reconcile.py
+++ b/adk/scripts/langfuse_gcp_reconcile.py
@@ -1,0 +1,441 @@
+#!/usr/bin/env python
+"""
+Reconcile Langfuse-reported LLM cost against the GCP Vertex AI bill.
+
+Langfuse computes cost from `usageDetails` keys, matched *exactly* against the price keys of a
+model definition, at ingestion time. Any token that lands in a key with no price — or on an
+observation whose model name never resolved — is silently free. That makes it possible for
+Langfuse to under-report real spend by ~2x while looking perfectly healthy.
+
+This script quantifies that. It pulls generations from the Langfuse API, re-prices them using the
+effective per-token rates *derived from the billing CSV itself* (rather than hardcoded list
+prices, so it survives repricing and new models), and prints where the two sides diverge.
+
+It also emits two health metrics that work as a standing regression check, no CSV needed:
+
+  * bucket mismatch — share of generations where the priced usage keys don't sum to `total`.
+    Every such token is billed by Google and free in Langfuse.
+  * unpriced models — generations whose model name never resolved, so the whole call cost $0.
+
+Usage:
+
+    # full reconciliation against a bill
+    uv run python scripts/langfuse_gcp_reconcile.py \
+        --billing-csv ~/Downloads/ocf-ikalatv-20260701-20260731.csv \
+        --from 2026-07-01 --to 2026-08-01
+
+    # health check only (no bill needed) — non-zero exit if thresholds are breached
+    uv run python scripts/langfuse_gcp_reconcile.py --from 2026-07-01 --to 2026-08-01 \
+        --max-bucket-mismatch 0.02 --max-unpriced-models 0
+
+Reads LANGFUSE_PUBLIC_KEY / LANGFUSE_SECRET_KEY / LANGFUSE_BASE_URL from the environment, the
+same variables `adk/instrumentation.py` uses. Note the API key is project-scoped, so this only
+sees one Langfuse project per run.
+"""
+
+from __future__ import annotations
+
+import argparse
+import csv
+import os
+import re
+import sys
+from collections import defaultdict
+from dataclasses import dataclass, field
+
+import httpx
+
+# Langfuse stores the OTel attribute `llm.token_count.completion_details.reasoning` under this
+# usage key. No Langfuse-managed Gemini model definition prices it (they price `output_reasoning`
+# / `thoughtsTokenCount`), so thinking tokens are free unless we map them ourselves.
+REASONING_KEY = "completion_details.reasoning"
+
+# Modality words that separate the family prefix from the direction in a Vertex SKU description,
+# e.g. "Gemini 3.1 Flash Lite Global | Video | Input - Predictions".
+_MODALITIES = ("text", "video", "audio", "image")
+
+# SKU noise that isn't part of the model family name.
+_SKU_NOISE = re.compile(r"\b(ga|global|thinking|predictions)\b|\(.*?\)")
+
+# SKUs that aren't per-token chat generation and have no Langfuse counterpart.
+_NON_CHAT_SKU = re.compile(r"embedding|grounding with google search", re.I)
+
+
+def canonical_family(name: str) -> str:
+    """
+    Normalize a Vertex SKU description or a Langfuse model id to a shared family key.
+
+        "Gemini 3.1 Flash Lite Global Video Input - Predictions" -> "gemini 3.1 flash lite"
+        "gemini-3.1-flash-lite-preview"                          -> "gemini 3.1 flash lite"
+    """
+    text = name.strip().lower().replace("-", " ").replace("_", " ")
+    text = text.split(" - ")[0]
+    # Cut the SKU at its modality word, which always follows the family name.
+    for modality in _MODALITIES:
+        idx = text.find(f" {modality} ")
+        if idx != -1:
+            text = text[:idx]
+            break
+    text = _SKU_NOISE.sub(" ", text)
+    # Version-suffixed and preview model ids belong to their base family.
+    text = re.sub(r"\b(preview|exp|latest|\d{3,})\b", " ", text)
+    return " ".join(text.split())
+
+
+def sku_kind(sku: str) -> str:
+    """Classify a SKU as cached input, plain input, or output."""
+    low = sku.lower()
+    if "caching" in low:
+        return "cached"
+    if "output" in low:
+        return "output"
+    return "input"
+
+
+@dataclass
+class Bucket:
+    tokens: int = 0
+    cost: float = 0.0
+
+
+@dataclass
+class Family:
+    """One model family's billed tokens and cost, split by direction."""
+
+    buckets: dict[str, Bucket] = field(default_factory=lambda: defaultdict(Bucket))
+
+    @property
+    def cost(self) -> float:
+        return sum(b.cost for b in self.buckets.values())
+
+    def rate(self, *kinds: str) -> float:
+        """Effective $/token across the given buckets — blends the cache discount in."""
+        tokens = sum(self.buckets[k].tokens for k in kinds if k in self.buckets)
+        cost = sum(self.buckets[k].cost for k in kinds if k in self.buckets)
+        return cost / tokens if tokens else 0.0
+
+
+@dataclass
+class Bill:
+    families: dict[str, Family]
+    line_item_total: float = 0.0
+    non_chat: float = 0.0
+    #: The export's own "Filtered total" row, which is authoritative — GCP's per-row
+    #: "Unrounded subtotal" values don't always sum to it exactly.
+    stated_total: float | None = None
+
+    @property
+    def chat(self) -> float:
+        return self.line_item_total - self.non_chat
+
+
+def load_billing(path: str) -> Bill:
+    """Parse a GCP billing export grouped by SKU into per-family buckets."""
+    bill = Bill(families=defaultdict(Family))
+    with open(path, newline="") as handle:
+        for row in csv.DictReader(handle):
+            sku = (row.get("SKU description") or "").strip()
+            try:
+                cost = float(row.get("Unrounded subtotal ($)") or 0)
+            except ValueError:
+                continue
+            if not sku:
+                # Trailing summary rows: the label sits in a shifted column.
+                if "filtered total" in " ".join(v or "" for v in row.values()).lower():
+                    bill.stated_total = cost
+                continue
+            try:
+                tokens = int((row.get("Usage amount") or "0").replace(",", ""))
+            except ValueError:
+                continue
+            bill.line_item_total += cost
+            if _NON_CHAT_SKU.search(sku):
+                bill.non_chat += cost
+                continue
+            bucket = bill.families[canonical_family(sku)].buckets[sku_kind(sku)]
+            bucket.tokens += tokens
+            bucket.cost += cost
+    return bill
+
+
+def fetch_generations(
+    base_url: str, auth: tuple[str, str], start: str, end: str
+) -> list[dict]:
+    """Page through every GENERATION observation in the window."""
+    out: list[dict] = []
+    with httpx.Client(base_url=base_url, auth=auth, timeout=120.0) as client:
+        page = 1
+        while True:
+            resp = client.get(
+                "/api/public/observations",
+                params={
+                    "type": "GENERATION",
+                    "fromStartTime": start,
+                    "toStartTime": end,
+                    "limit": 100,
+                    "page": page,
+                },
+            )
+            resp.raise_for_status()
+            body = resp.json()
+            batch = body.get("data") or []
+            out.extend(batch)
+            meta = body.get("meta") or {}
+            print(
+                f"  page {page}/{meta.get('totalPages', '?')} ({len(out)} so far)",
+                file=sys.stderr,
+            )
+            if page >= (meta.get("totalPages") or 0) or not batch:
+                return out
+            page += 1
+
+
+@dataclass
+class Split:
+    """A generation's tokens, split into what Langfuse priced and what it silently dropped."""
+
+    input_priced: int = 0
+    input_tool_use: int = 0
+    output_priced: int = 0
+    output_reasoning: int = 0
+    bucket_mismatch: bool = False
+
+    @property
+    def input_total(self) -> int:
+        return self.input_priced + self.input_tool_use
+
+    @property
+    def output_total(self) -> int:
+        return self.output_priced + self.output_reasoning
+
+
+def split_usage(usage: dict) -> Split:
+    """
+    Separate a generation's tokens into priced and unpriced.
+
+    Two span shapes exist in practice. Fully-attributed spans set `output` from
+    `gen_ai.usage.output_tokens` (candidates only), so reasoning is *additive* and the leftover
+    `total - input - output - reasoning` is `toolUsePromptTokenCount` — a field Gemini reports
+    separately from `promptTokenCount`, bills at the input rate, and that no OTel attribute
+    carries. Truncated spans (no `gen_ai.*` attributes) instead set `output` from
+    `llm.token_count.completion`, which already includes reasoning; there, input + output == total
+    and nothing is missing.
+    """
+    total = usage.get("total") or 0
+    inp = usage.get("input") or 0
+    out = usage.get("output") or 0
+    reasoning = usage.get(REASONING_KEY) or 0
+
+    if inp + out == total:
+        # Truncated-span shape: `output` already subsumes reasoning, nothing unaccounted for.
+        return Split(input_priced=inp, output_priced=out)
+
+    leftover = total - inp - out - reasoning
+    return Split(
+        input_priced=inp,
+        input_tool_use=max(leftover, 0),
+        output_priced=out,
+        output_reasoning=reasoning,
+        bucket_mismatch=True,
+    )
+
+
+def main() -> int:
+    parser = argparse.ArgumentParser(
+        description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter
+    )
+    parser.add_argument(
+        "--billing-csv",
+        help="GCP billing export grouped by SKU. Omit for health check only.",
+    )
+    parser.add_argument(
+        "--from", dest="start", required=True, help="inclusive start date, YYYY-MM-DD"
+    )
+    parser.add_argument(
+        "--to", dest="end", required=True, help="exclusive end date, YYYY-MM-DD"
+    )
+    parser.add_argument(
+        "--fallback-family",
+        default="gemini 3 flash",
+        help="family used to price generations whose model never resolved (default: %(default)s)",
+    )
+    parser.add_argument(
+        "--max-bucket-mismatch",
+        type=float,
+        help="fail if the mismatch share exceeds this (0-1)",
+    )
+    parser.add_argument(
+        "--max-unpriced-models",
+        type=int,
+        help="fail if more than this many generations lack a model",
+    )
+    args = parser.parse_args()
+
+    public, secret = os.getenv("LANGFUSE_PUBLIC_KEY"), os.getenv("LANGFUSE_SECRET_KEY")
+    base_url = os.getenv("LANGFUSE_BASE_URL") or os.getenv("LANGFUSE_HOST")
+    if not (public and secret and base_url):
+        print(
+            "Set LANGFUSE_PUBLIC_KEY, LANGFUSE_SECRET_KEY and LANGFUSE_BASE_URL.",
+            file=sys.stderr,
+        )
+        return 2
+
+    start, end = f"{args.start}T00:00:00Z", f"{args.end}T00:00:00Z"
+    print(
+        f"Fetching generations {args.start} .. {args.end} from {base_url}",
+        file=sys.stderr,
+    )
+    generations = fetch_generations(base_url, (public, secret), start, end)
+    if not generations:
+        print("No generations in window.", file=sys.stderr)
+        return 0
+
+    # ---- Langfuse side -------------------------------------------------------------------
+    per_family: dict[str, Split] = defaultdict(Split)
+    per_agent: dict[str, Split] = defaultdict(Split)
+    reported = 0.0
+    unpriced_models = 0
+    mismatches = 0
+
+    for obs in generations:
+        split = split_usage(obs.get("usageDetails") or {})
+        reported += obs.get("calculatedTotalCost") or 0.0
+        mismatches += split.bucket_mismatch
+
+        model = obs.get("model")
+        if model:
+            family = canonical_family(model)
+        else:
+            family = args.fallback_family
+            unpriced_models += 1
+
+        agent = ((obs.get("metadata") or {}).get("attributes") or {}).get(
+            "gen_ai.agent.name"
+        ) or "<unattributed>"
+        for key, acc in ((family, per_family), (agent, per_agent)):
+            target = acc[key]
+            target.input_priced += split.input_priced
+            target.input_tool_use += split.input_tool_use
+            target.output_priced += split.output_priced
+            target.output_reasoning += split.output_reasoning
+
+    n = len(generations)
+    print(
+        f"\n{'=' * 92}\nLANGFUSE HEALTH — {n:,} generations, {args.start} .. {args.end}\n{'=' * 92}"
+    )
+    print(f"  reported cost                                   ${reported:>10.3f}")
+    print(
+        f"  bucket mismatch (priced keys != total)  {mismatches:>6,} / {n:,}  ({mismatches / n * 100:.1f}%)"
+    )
+    print(
+        f"  model never resolved -> whole call $0    {unpriced_models:>6,} / {n:,}  ({unpriced_models / n * 100:.1f}%)"
+    )
+
+    print("\n  tokens Langfuse holds but does not price, by agent:")
+    print(
+        f"    {'agent':<26}{'input priced':>14}{'TOOL-USE':>12}{'output':>10}{'REASONING':>12}"
+    )
+    for agent, s in sorted(
+        per_agent.items(), key=lambda kv: -kv[1].input_tool_use - kv[1].output_reasoning
+    ):
+        print(
+            f"    {agent:<26}{s.input_priced:>14,}{s.input_tool_use:>12,}"
+            f"{s.output_priced:>10,}{s.output_reasoning:>12,}"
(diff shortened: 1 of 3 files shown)


MrOrz:
## Live A/B on the preview deploy: both invariants hold, and the ADR's load-bearing assumption is confirmed on real traffic

Six fact-check sessions run on 2026-08-29, same Langfuse project, differing only in which build served them:

- **control** — `dev.cofacts.ai` (staging, `master`): `6953a34a`, `19df2871`, `c1bac733`
- **test** — PR 122 preview deploy: `aaa732d6`, `68400485`, `dd3bdb84`

`scripts/langfuse_check_usage.py --from 2026-08-29 --to 2026-08-30`, one environment each (the window contains exactly these six sessions):

| | control (`staging`) | test (`preview`) |
| --- | ---: | ---: |
| generations | 75 | 63 |
| tokens in `total` | 3,786,708 | 1,761,837 |
| **unpriced tokens** | **446,573 (11.8%)** | **0 (0.0%)** |
| **token-carrying generations at $0** | **10 / 75** | **0 / 63** |
| reported cost | $1.072 | $0.762 |

Both thresholds pass on the test group: `--max-unpriced-share 0.02 --max-unpriced-generations 0` exits 0 on `preview` and exits 1 on `staging`.

Usage keys actually stored, aggregated over each group:

| group | keys on every generation |
| --- | --- |
| control | `input`, `output`, `completion_details.reasoning` (+ `prompt_details.audio` on 2) |
| test | `input`, `input_cached_tokens`, `output`, `output_reasoning` |

and the per-key `costDetails` confirm each of those resolves to a price on the managed definitions — `gemini-3-flash-preview` at $0.50/$3.00 per Mtok with `input_cached_tokens` at $0.05 (the 90% discount, applied for the first time), `gemini-3.1-flash-lite` at $0.25/$1.50, `output_reasoning` priced at the output rate in both.

### The ADR's open concerns, one by one

**"Depends on `after_model_callback` running inside the instrumentor's `call_llm` span" — confirmed on live traffic, not just under test.** All **63/63** test generations carry the instrumentor's `llm.token_count.*` *and* our `langfuse.observation.usage_details` / `langfuse.observation.model.name` on the **same observation**. No second observation is created anywhere in the six traces, and the observation mix per trace is unchanged in shape (25 GENERATION / 15 AGENT / 20 TOOL / 15 CHAIN on `aaa732d6`). The fallback of writing to the span directly stays unnecessary.

**"Overwrite, not patch, is load-bearing on 0.1.24."** Holds: no test generation retains a `completion_details.*` or `prompt_details.*` key, so the instrumentor's own (now tool-use-folded) values were replaced wholesale rather than added to. The arithmetic proves there is no double count either — `sum(priced keys) == total` **exactly on 63/63** test generations, and the stored `total` equals the instrumentor's independent `llm.token_count.total` to the token in both groups (1,761,837 / 3,786,708). Control, by contrast, is exact on only 55/75: 13 overshoot (`completion` already includes reasoning, so `completion_details.reasoning` double-counts — harmless only because it was unpriced) and 7 undershoot (the tool-use residue).

**"`model = null` on streamed root-agent spans" (cause 2).** 9 of the 63 test generations still arrive with **no `llm.model_name` from the instrumentor** — the bug is unchanged upstream, as expected — yet every one of them resolved a model and a non-zero cost, via the `ContextVar` carried from `before_model_callback`. Those same spans are 10/75 at $0 in control, covering 1.21M tokens (~$0.64 of the sessions' true cost given away).

**"Streamed callback fires once per chunk; the `usage_metadata is None` guard must not clobber."** No test generation has `total == 0`, and every one has a full four-key breakdown — so on streamed turns the last usage-bearing write survived and no empty aggregated response overwrote it.

**Cached tokens (cause 5)** are now emitted: 828,771 of them on `gemini-3-flash-preview`. Under `master` these would have been billed inside `input` at full rate — $0.414 instead of $0.041 on this sample.

### Size of the gap, on comparable traffic

The two groups ran different conversations, so their totals are not directly comparable. Re-pricing the **control** usage at the rates the test group's `costDetails` reveal — everything that is not output (tool-use residue, audio, cached, tool-use folded input) at the input rate, reasoning at the output rate except where `completion` already contains it, and no cache discount since `master` never emits that bucket:

| session | Langfuse reported | re-priced | ratio |
| --- | ---: | ---: | ---: |
| `6953a34a` | $0.2248 | $0.5635 | 2.51x |
| `19df2871` | $0.7634 | $1.3550 | 1.78x |
| `c1bac733` | $0.0841 | $0.2345 | 2.79x |
| **total** | **$1.0722** | **$2.1530** | **2.01x** |

which lands on the ~2x the ADR derived from the July and August billing reconciliations, from an entirely independent direction. (The re-price is an estimate, not a bill: it cannot know the audio premium or how much of the residue Google charged at the cached rate.)

### One residual worth noting (not a blocker)

Two control generations carry `prompt_details.audio` (50,417 tokens), priced at **$0** because the managed `gemini-3-flash-preview` definition has no audio key at all. After this PR those tokens fold into `input` and are charged at the text rate — strictly better than free, but `gemini-3.1-flash-lite`'s managed definition prices `input_audio_tokens` at 2x its text rate, so if Google bills `gemini-3-flash-preview` audio at a premium too we would still under-report that slice. `usage_metadata.prompt_tokens_details` carries the per-modality split, so emitting `input_audio_tokens` is a small follow-up if audio input becomes a material share.

Verdict: the implementation does what the ADR says it should, on live streamed traffic, and the assumption the ADR flagged as "must be verified during implementation" is verified.

---

adk/instrumentation.py
```diff
@@ -100,3 +115,74 @@ async def before_run_callback(self, *, invocation_context: InvocationContext):
...
+            # place is the right outcome there.
+            return None
+
+        prompt = usage.prompt_token_count or 0
+        # `prompt_token_count` *includes* `cached_content_token_count`, so the two
+        # are split rather than added, or cached tokens get counted twice.
+        cached = usage.cached_content_token_count or 0
+
+        get_client().update_current_generation(
+            model=llm_response.model_version or _request_model.get(),
```
MrOrz: **Preferring `model_version` over the request model can break price matching that works today.**

The instrumentor sets `llm.model_name` from `llm_request.model` (`openinference/instrumentation/google_adk/_wrappers.py:296-297`) — the undated agent model id (`gemini-3-flash-preview`), which is the exact spelling the ADR verified carries managed prices. `langfuse.observation.model.name` takes precedence at ingestion, so this line replaces it on *every* generation, not just the ones that were missing a model.

Vertex returns `modelVersion` date-suffixed — the test's own example is `gemini-3-flash-preview-11-2026`. If a managed definition's `matchPattern` doesn't cover the suffix, `modelId` resolves to null and `costDetails` is empty: non-streamed generations that price correctly today flip to $0, permanently, since cost is fixed at ingestion and never backfilled. That is the same silent-$0 failure the ADR documents for `gemini-3.5-flash`.

Note the ADR's own proposed shape writes `model=<the agent's model id>`, i.e. the request side; the implementation inverted the preference without recording why. Flipping the order still repairs the `model = null` streamed spans (`model_version` is unset there anyway):

```suggestion
            model=_request_model.get() or llm_response.model_version,
```

If the dated name is preferred deliberately, worth confirming on the instance that `gemini-3-flash-preview-11-2026` resolves to a model id before merging.

---

adk/scripts/langfuse_check_usage.py
```diff
@@ -0,0 +1,199 @@
...
+def unpriced_tokens(usage: dict) -> int:
+    """Tokens inside `total` that no priced key accounts for.
+
+    `total` is Langfuse's derived aggregate — "not a bucket itself but spans all buckets and
+    equals their sum" — so any shortfall is a bucket that was never sent, or was sent under a
+    key with no price.
+    """
+    total = usage.get("total") or 0
+    priced = sum(v for k, v in usage.items() if k != "total" and isinstance(v, int))
+    return max(total - priced, 0)
```
MrOrz: **This sums *all* keys, not the priced ones — so an unpriced key hides itself.**

The docstring (and the ADR) state the invariant as "priced keys must sum to `total`", but nothing here consults pricing: every integer key counts toward `priced`, whether or not Langfuse had a price for it.

Concrete miss: a Gemini family arrives whose managed definition has no `input_cached_tokens` / `output_reasoning` price (exactly the `gemini-3.5-flash` model-list lag documented in the ADR), or Langfuse renames those keys. The plugin keeps emitting them, Google bills those tokens, Langfuse gives them away — and this returns 0 for every observation, so the run prints `0.0%` and exits 0. The second invariant doesn't catch it either: `input`/`output` still price, so `calculatedTotalCost > 0`.

There's also a structural problem now that the fix is in. The plugin sends `usage_details` with no `total`, and Langfuse derives `total` as the sum of the non-total keys it received ("the total key is the sum of all (non-total) usage metrics *or the total value ingested*"). On the fixed path the shortfall is therefore 0 by construction — the only regression this metric can still detect is "our write stopped landing at all", not the key-naming class of bug it was written for (causes 3 and 5).

`costDetails` comes back from the same endpoint and names exactly the keys that got a price, so the stated invariant is directly testable:

```python
def unpriced_tokens(usage: dict, costs: dict) -> int:
    """Tokens under a usage key that Langfuse attached no price to."""
    priced_keys = {k for k in costs if k != "total"}
    unpriced = sum(
        v for k, v in usage.items()
        if k != "total" and isinstance(v, int) and k not in priced_keys
    )
    # plus any bucket that never arrived at all
    return unpriced + max((usage.get("total") or 0) - sum(
        v for k, v in usage.items() if k != "total" and isinstance(v, int)
    ), 0)
```

---

adk/scripts/langfuse_check_usage.py
```diff
@@ -0,0 +1,199 @@
...
+
+def unpriced_tokens(usage: dict) -> int:
+    """Tokens inside `total` that no priced key accounts for.
+
+    `total` is Langfuse's derived aggregate — "not a bucket itself but spans all buckets and
+    equals their sum" — so any shortfall is a bucket that was never sent, or was sent under a
+    key with no price.
+    """
+    total = usage.get("total") or 0
+    priced = sum(v for k, v in usage.items() if k != "total" and isinstance(v, int))
```
MrOrz: This treats *every* non-`total` key as priced, so a bucket Langfuse receives but cannot price counts as accounted for — which is exactly cause 3, the one this check is meant to stand guard over.

The ADR's own July record shows it: Langfuse stored `{input: 1287, output: 729, completion_details.reasoning: 9963, total: 2955369}`. The 9963 reasoning tokens carry no price in any managed Gemini definition, yet they land in `priced`, so only the tool-use residue is reported.

Concrete regression the check would wave through: an ADK/instrumentor upgrade stops `after_model_callback` from firing (the very thing this script exists to catch) while instrumentor ≥0.1.18 folds tool-use into `prompt`. Stored keys become `{input: 2944677, output: 729, completion_details.reasoning: 9963, total: 2955369}` — sum equals `total`, `unpriced_tokens` returns 0, `--max-unpriced-share 0.02` exits 0, and the thinking tokens Google bills are free in Langfuse again.

Suggested shape: sum only the keys we know carry a price and count everything else as unpriced.

```suggestion
    total = usage.get("total") or 0
    priced = sum(
        v
        for k, v in usage.items()
        if k in _PRICED_KEYS and isinstance(v, int)
    )
```
(with `_PRICED_KEYS = {"input", "input_cached_tokens", "output", "output_reasoning"}` — the four the plugin emits.)

---

adk/scripts/langfuse_check_usage.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def unpriced_tokens(usage: dict) -> int:
+    """Tokens inside `total` that no priced key accounts for.
+
+    `total` is Langfuse's derived aggregate — "not a bucket itself but spans all buckets and
+    equals their sum" — so any shortfall is a bucket that was never sent, or was sent under a
+    key with no price.
+    """
+    total = usage.get("total") or 0
```
MrOrz: Worth verifying before this is trusted in CI: after the fix, is `total` still an *independent* reference?

Langfuse defines it as "the sum of all (non-total) usage metrics **or the total value ingested**" (`langfuse/api/commons/types/observation.py:121-126`). Pre-fix the instrumentor ingested an explicit `llm.token_count.total`, which is precisely what made the 25.6% shortfall visible. Post-fix, `langfuse.observation.usage_details` takes precedence and carries no `total` key, so the stored total is very likely computed as the sum of our own four keys — in which case `total - priced` is identically 0 for every generation the plugin touched, and invariant 1 can never fail again.

Failure that would slip through: a later edit drops or misspells `output_reasoning`. Those tokens disappear from `usage_details` *and* from the derived `total`, `unpriced_tokens()` returns 0, `--max-unpriced-share` passes, and thinking tokens go unbilled in Langfuse while Google bills them.

If a real post-deploy generation confirms the derived total, the reference has to come from somewhere the plugin does not overwrite — e.g. also send `"total": usage.total_token_count` from `after_model_callback`, so a missing bucket shows up as a genuine shortfall rather than shrinking the yardstick with it.

---

MrOrz (commented):
Reviewed the usage-mapping shim, the regression script and the tests.

The plugin arithmetic checks out against Gemini's `usage_metadata` semantics (`prompt_token_count` includes cached, `tool_use_prompt_token_count` is separate and inside `total_token_count`, `candidates_token_count` excludes thoughts), the `usage is None` guard behaves as documented in SSE, and I confirmed in ADK 1.26 that `_handle_after_model_callback` runs inside the `call_llm` span (`base_llm_flow.py:1124-1180`) and that `llm_request.model` is populated by `flows/llm_flows/basic.py:47` before `before_model_callback` fires — so both the span target and the model fallback hold. Plugins also propagate into `AgentTool` sub-runners (`include_plugins=True` by default), so sub-agent generations get mapped too. All 84 tests pass.

Two findings, both on the standing check rather than the fix itself — the check is the part claimed to be version-independent, and I don't think it currently is.

---

MrOrz:
## Re-run on `f447fa6`: same result, now from a check that could have failed

Reviewing the diff after the A/B above turned up three ways the fix could be wrong on traffic that run happened not to produce — one of them in the measuring instrument, which makes part of that first comment weaker than it reads. All three are fixed in `f447fa6`, and three fresh fact-check sessions through the redeployed preview re-measure it:

| | control (`staging`, `master`) | test (`preview`, `f447fa6`) |
| --- | ---: | ---: |
| generations | 75 | 71 |
| tokens in `total` | 3,786,708 | 1,902,335 |
| **tokens Langfuse priced nothing for** | **1,802,168 (47.6%)** | **0 (0.0%)** |
| **token-carrying generations at $0** | **10 / 75** | **0 / 71** |
| reported cost | $1.072 | $0.933 |

`scripts/langfuse_check_usage.py --environment preview --from 2026-08-29 --to 2026-08-30 --max-unpriced-share 0.02 --max-unpriced-generations 0` exits 0 — over all 134 preview generations of the day, both rounds — and the same thresholds against `--environment staging` exit 1.

### What was wrong, and what it changes above

**`total` was derived, so the first invariant could never fail.** The plugin sent four buckets and no `total`; Langfuse derives an absent `total` by summing the keys it received. The yardstick was therefore computed from the very buckets it was meant to check — a shortfall of 0 was guaranteed on every generation the plugin touched, and a bucket a later edit dropped or misspelled would have vanished from the total along with the tokens. So this line of my earlier comment does not hold, and I am retracting it:

> The arithmetic proves there is no double count either — `sum(priced keys) == total` **exactly on 63/63** test generations

The rest of that paragraph does hold: the instrumentor's own `llm.token_count.total` stays on the span untouched by our overwrite, and matching it was — and remains — the non-circular half of the evidence.

The plugin now ingests `usage_metadata.total_token_count`, which is visible on the span itself:

```json
{"input": 5433, "input_cached_tokens": 38860, "output": 216, "output_reasoning": 57, "total": 44566}
```

All **71/71** test generations carry an ingested `total`, and our four buckets sum to it exactly on all 71 — this time a comparison against Gemini's own count rather than an identity. The independent cross-check agrees: summed over the group, the instrumentor's `llm.token_count.total` is the same 1,902,335.

**The check script counted every key as priced.** `unpriced_tokens` summed all non-`total` keys, which catches a bucket that never arrived (cause 1, tool-use tokens hiding inside `total`) but not a bucket that arrived under a name carrying no price (cause 3, `completion_details.reasoning`). Those are the two distinct ways a token goes free and only one was being measured. It now decides by `costDetails`, which names the keys that actually earned a price. That is why the control column reads **47.6%** here against 11.8% in the first comment — same sessions, same data, stricter and correct metric. It also moves the August pre-fix baseline to **56.9% of 33.9M tokens** (from 25.2%), which is much closer to the 56% of real July spend the ADR found missing against the bill; the ADR's Confirmation section is updated to match, and `<unresolved>` now heads the by-model breakdown instead of hiding in the second invariant.

**The model id could have been a dated one.** The implementation preferred `llm_response.model_version`, and Vertex can answer with a dated build such as `gemini-3-flash-preview-11-2026`. On this instance every managed Gemini definition from 2.5 onwards matches the undated id and ends in a hard `$` — the optional group the older ones carry is Vertex's `@version` form, not a date suffix — so a dated name matches nothing, resolves to no model, and costs $0 permanently: exactly the `gemini-3.5-flash` failure the ADR documents. It never fired in either round (all names came out undated), but nothing in the code prevented it. The request-side id is preferred now, `model_version` is the fallback, and the streamed responses that needed repairing leave it unset anyway.

### The rest of the evidence, restated on this round

- **The span-targeting assumption** — all 71/71 test generations carry the instrumentor's `llm.token_count.*` and our `langfuse.observation.*` on the **same observation**; no stray second observation in any of the three traces.
- **Cause 2** — 13 of 71 still arrive with no `llm.model_name` from the instrumentor, and all 13 resolved a model and a non-zero cost from the `ContextVar`. In control the same spans are 10/75 at $0, 1.21M tokens given away.
- **Cause 5** — 699,872 cached tokens billed $0.035, where `master` charges $0.350 for them inside `input`.
- **Keys and prices** — `input`, `input_cached_tokens`, `output`, `output_reasoning` on every generation, each with its own `costDetails` entry, so all four resolved to a real price.
- **The streamed-clobber guard** — no test generation has `total == 0`, and every one carries the full breakdown.
- **Size of the gap** — unchanged, since it is computed from the control sessions: re-pricing them at the test group's revealed rates gives $1.0722 → $2.1530, **2.01x**.

`uv run pytest` is at 89 passed, including five new tests pinning the check script's own arithmetic — the regression that hid here is now itself guarded.

Still open, and not a blocker: `gemini-3-flash-preview`'s managed definition has no audio key, so the 50,417 `prompt_details.audio` tokens that control priced at $0 now fold into `input` at the text rate. Better than free, but `gemini-3.1-flash-lite` prices `input_audio_tokens` at twice its text rate, so a premium on flash audio would still be under-reported. `usage_metadata.prompt_tokens_details` carries the per-modality split if audio input ever becomes a material share.

---

adk/instrumentation.py
MrOrz: Flipped in `f447fa6` — `model=_request_model.get() or llm_response.model_version`, with the reason recorded next to it and in the ADR.

Checked the instance rather than assuming: all 25 managed Gemini definitions match the undated id, and from 2.5 onwards they end in a hard `$` (`(?i)^(google(ai)?/)?(gemini-3-flash-preview)$`). The optional group the 1.x/2.0 ones carry is `(@[a-zA-Z0-9]+)?`, i.e. Vertex's `@version` form, not a date suffix — dated builds get their own definition instead (`gemini-2.0-flash-lite-preview-02-05`). So a dated name would have matched nothing.

It never fired on live traffic: across both preview rounds every `langfuse.observation.model.name` came out undated and every generation resolved a `modelId` with a non-zero cost. Latent rather than active, but nothing in the code prevented it.

`test_model_comes_from_the_request_then_the_response` now pins the order, including that a dated `model_version` does not win over the request id.

---

adk/scripts/langfuse_check_usage.py
MrOrz: Fixed in `f447fa6`. `unpriced_tokens(usage, costs)` now takes the observation's `costDetails` and counts the two losses separately: tokens under a key that earned no price, plus the shortfall of `total` over everything received.

The effect on real data is large — the same control sessions read **47.6%** unpriced instead of 11.8%, and the August pre-fix window **56.9% of 33.9M tokens** instead of 25.2%, which is much closer to the 56% of July spend the ADR found missing against the bill. `<unresolved>` now heads the by-model breakdown instead of hiding in the second invariant.

`cofacts_ai/tests/test_langfuse_check_usage.py` pins both loss shapes, an unresolved model, and the fixed-path shape.

---

adk/scripts/langfuse_check_usage.py
MrOrz: Confirmed, then fixed in `f447fa6`. Langfuse's docs are explicit — `"total": int, # if not set, it is derived as the sum of all usage types` — so on the fixed path the shortfall was 0 by construction and the invariant could not fail.

The plugin now sends `"total": usage.total_token_count`, which is visible on the span itself:

```json
{"input": 5433, "input_cached_tokens": 38860, "output": 216, "output_reasoning": 57, "total": 44566}
```

On the re-run through the redeployed preview, all 71/71 generations carry an ingested `total` and the four buckets sum to it exactly — a comparison against Gemini's own count now rather than an identity. The instrumentor's untouched `llm.token_count.total` agrees at 1,902,335 for the group.

This also retires the circular half of the evidence in my first comment on this PR, retracted there.

---

adk/scripts/langfuse_check_usage.py
MrOrz: Superseded — this one raised both script issues together, and each is handled on its own thread below (priced-key detection at `:95`, derived `total` at `:94`). Both are fixed in `f447fa6`; resolving here to avoid a third copy of the same discussion.

---

bil4444 (approved):

PR #123, opened 2026-08-04
feat(deploy): run url-resolver as a Cloud Run sidecar

MrOrz:
Stacked on #118 — that PR adds the url-resolver client, this one gives it something to talk to on Cloud Run. Base will need retargeting to `master` once #118 merges.

## Problem

`URL_RESOLVER_ADDRESS` defaults to `url-resolver:4000`, a docker-compose service name that does not resolve on Cloud Run. Deployed as-is, the verifier's page pre-fetch would take the `RESOLVER_UNAVAILABLE` path on every turn and quietly fall back to `url_context`-only — exactly the behaviour #118 exists to replace, with no loud failure to notice it by.

## Change

A fourth container in `service.template.yaml`, mirroring how rumors-deploy's `docker-compose.sample.yml` runs it:

| | |
|---|---|
| image | `docker.io/cofacts/url-resolver:latest` (public Docker Hub, [supported by Cloud Run](https://docs.cloud.google.com/run/docs/deploying); cached up to an hour) |
| port | 4000, `PORT=4000` |
| resources | 1 vCPU / 2048 MiB |
| probe | TCP on 4000 |

Plus: `backend` gains `URL_RESOLVER_ADDRESS=localhost:4000` and a `container-dependencies` entry on `url-resolver`, and `docs/index.md` now describes four containers instead of three.

**Containment.** url-resolver is insecure gRPC with no auth. As a sidecar it is reachable only over the instance's localhost — the same property the compose deployment gets from its private network, and the reason it is not a separate Cloud Run service.

**Sizing.** 2 GiB is deliberate headroom, not a measured requirement: Cloud Run terminates the *whole instance* when any single container exceeds its memory limit, which on a 900 s request timeout means killing an agent turn in flight. Idle footprint measured at ~150 MiB (below); `SCRAPE_MAX_CONCURRENCY` (default 3) is the knob to lower before raising this limit.

Instance totals go from 2.5 vCPU / 3.25 GiB to 3.5 vCPU / 5.25 GiB.

## Verification

**Deployed.** Revision reached Ready with all four containers and the new totals — so Cloud Run accepted the fractional CPU sum, pulled the Docker Hub image, and the sidecar passed its startup probe. Preview serves HTTP 200. CI green.

Before that, locally: rendered `service.template.yaml` through the workflow's `envsubst` substitution and parsed the result (valid YAML, every `container-dependencies` reference resolves to a real container name), then ran the published image under a 1 vCPU / 1 GiB container limit:

```
gRPC server listening on 4000
Browser launched successfully.
```

8 chromium processes up, 153 MiB resident at idle. An end-to-end scrape could not be exercised in that sandbox — egress only goes through a TLS-intercepting proxy, so fetches fail at `node-fetch` before reaching puppeteer.

## Open risk: local puppeteer on Cloud Run

**Not yet confirmed working on Cloud Run**, and the deploy passing is not evidence either way. Chromium launches at module load and logs `Browser launched successfully.`; on failure it logs to Rollbar (disabled — no token configured) and leaves the gRPC server up. The startup probe is TCP, so it passes in both cases. A chromium that cannot start would surface only as `UNKNOWN_SCRAPE_ERROR` on every resolve, which the client buckets as "no signal" and degrades past silently.

Confirming it needs the sidecar's runtime log:

```
gcloud logging read \
  'resource.type=cloud_run_revision AND resource.labels.service_name=cofacts-ai
   AND labels."run.googleapis.com/container_name"=url-resolver' \
  --limit 50 --region asia-east1
```

The second risk is CPU throttling: the browser is one long-lived process reused across requests and frozen between them. `scrape.js` relaunches on `disconnected`, so a dropped CDP connection should self-heal, but the first affected request still fails.

If either bites, url-resolver already ships a Cloudflare Browser Rendering backend (`BROWSER_BACKEND=cloudflare` + `CLOUDFLARE_ACCOUNT_ID` / `CLOUDFLARE_API_TOKEN`) that moves chromium off the instance. Switching is env-only; not wired up here.

### DIFF
--- a/service.template.yaml
+++ b/service.template.yaml
@@ -11,7 +11,7 @@ spec:
     metadata:
       annotations:
         run.googleapis.com/execution-environment: gen2
-        run.googleapis.com/container-dependencies: '{"ingress":["backend"],"backend":["cloudsql-proxy"]}'
+        run.googleapis.com/container-dependencies: '{"ingress":["backend"],"backend":["cloudsql-proxy","url-resolver"]}'
         autoscaling.knative.dev/minScale: "0"
         autoscaling.knative.dev/maxScale: "3"
     spec:
@@ -75,6 +75,11 @@ spec:
           value: "${GCS_ARTIFACT_BUCKET}"
         - name: COFACTS_API_URL
           value: "${COFACTS_API_URL}"
+        # url-resolver runs as a sidecar in this same instance (see below), so
+        # the verifier's page pre-fetch talks to it over localhost instead of
+        # the docker-compose service name the client defaults to.
+        - name: URL_RESOLVER_ADDRESS
+          value: "localhost:4000"
         startupProbe:
           tcpSocket:
             port: 8000
@@ -97,6 +102,43 @@ spec:
         startupProbe:
           tcpSocket:
             port: 5432
+          initialDelaySeconds: 2
+          timeoutSeconds: 5
+          periodSeconds: 10
+          failureThreshold: 3
+      # Headless-browser page fetcher behind the verifier's URL pre-fetch. It
+      # speaks insecure gRPC with no auth, so it is never exposed as its own
+      # service: as a sidecar it is reachable only over this instance's
+      # localhost, the same containment the docker-compose deployment gets from
+      # its private network (rumors-deploy `docker-compose.sample.yml`).
+      - name: url-resolver
+        # Public Docker Hub image, published by cofacts/url-resolver on release
+        # tags (`:latest`) and master pushes (`:dev`). Cloud Run caches public
+        # Docker Hub images for up to an hour, so a freshly pushed `:latest` can
+        # take that long to appear in a new revision.
+        image: docker.io/cofacts/url-resolver:latest
+        resources:
+          limits:
+            # Chromium is what sizes this container: one browser process tree
+            # plus up to SCRAPE_MAX_CONCURRENCY (default 3) live pages. Cloud
+            # Run terminates the whole instance when any single container
+            # exceeds its memory limit — which would kill an agent turn in
+            # flight — so this is sized with headroom over the ~150 MiB idle
+            # footprint rather than trimmed to it. If it still OOMs, lower
+            # SCRAPE_MAX_CONCURRENCY before raising the limit.
+            cpu: "1000m"
+            memory: "2048Mi"
+        env:
+        - name: PORT
+          value: "4000"
+        startupProbe:
+          # The gRPC server binds immediately; chromium launches in the
+          # background right after. A TCP probe therefore only proves the RPC
+          # endpoint is up, not that a browser is ready — the client already
+          # treats an unreachable/failing resolver as "no signal" and falls
+          # back to `url_context`, so that is the right amount of gating here.
+          tcpSocket:
+            port: 4000
           initialDelaySeconds: 2
           timeoutSeconds: 5
           periodSeconds: 10
--- a/service.template.yaml
+++ b/service.template.yaml
@@ -126,11 +126,32 @@ spec:
             # flight — so this is sized with headroom over the ~150 MiB idle
             # footprint rather than trimmed to it. If it still OOMs, lower
             # SCRAPE_MAX_CONCURRENCY before raising the limit.
+            #
+            # This sizing follows the *local* browser path even when
+            # BROWSER_BACKEND is `cloudflare` and no chromium ever starts here.
+            # The backend is a repo variable that can revert to local at any
+            # moment, and a container trimmed to the Cloudflare footprint would
+            # OOM the instance the moment someone flips it back. Trim these only
+            # once the local fallback is retired for good.
             cpu: "1000m"
             memory: "2048Mi"
         env:
         - name: PORT
           value: "4000"
+        # Where puppeteer's browser runs. Unset renders as "" here, which
+        # url-resolver reads as `local` (`process.env.BROWSER_BACKEND || 'local'`),
+        # so an unconfigured deploy keeps launching chromium inside this
+        # container. Set the URL_RESOLVER_BROWSER_BACKEND repo variable to
+        # `cloudflare` to move it to Cloudflare Browser Rendering instead; clear
+        # the variable to roll back.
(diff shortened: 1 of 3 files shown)


service.template.yaml
```diff
@@ -11,7 +11,7 @@ spec:
     metadata:
       annotations:
         run.googleapis.com/execution-environment: gen2
-        run.googleapis.com/container-dependencies: '{"ingress":["backend"],"backend":["cloudsql-proxy"]}'
+        run.googleapis.com/container-dependencies: '{"ingress":["backend"],"backend":["cloudsql-proxy","url-resolver"]}'
```
MrOrz: Adding `url-resolver` to `backend`'s `container-dependencies` couples the whole revision's availability to the sidecar: if it fails its startup probe, `backend` never starts and the revision is never Ready — the fact-checking service is down, not degraded.

That contradicts the client's deliberate design in #118, where an unreachable resolver buckets as `RESOLVER_UNAVAILABLE` ("no signal") and falls back to `url_context`. The risk is concrete because the image is the mutable `docker.io/cofacts/url-resolver:latest`: an upstream push that breaks startup takes the site down on the next cold start, with no deploy on this side to correlate it with.

Since the probe is TCP-only and the gRPC server binds immediately (as the PR body notes), the dependency also buys very little — it doesn't prove a browser is reachable. Suggest dropping `url-resolver` from the dependency list and keeping only `cloudsql-proxy`, letting the resolver's absence degrade the way the client already handles.

---

service.template.yaml
```diff
@@ -75,6 +75,11 @@ spec:
           value: "${GCS_ARTIFACT_BUCKET}"
         - name: COFACTS_API_URL
           value: "${COFACTS_API_URL}"
+        # url-resolver runs as a sidecar in this same instance (see below), so
+        # the verifier's page pre-fetch talks to it over localhost instead of
+        # the docker-compose service name the client defaults to.
+        - name: URL_RESOLVER_ADDRESS
```
MrOrz, line 89: Companion to the client-side finding on #118 (`url_resolver/client.py:28`): `URL_RESOLVER_TIMEOUT` isn't set alongside the address, so production runs on the client's 30s default — a deadline for the *entire* streaming call covering up to 20 URLs at concurrency 3, now against a remote Cloudflare browser rather than an in-process chromium. On a message with several links, most URLs realistically come back `TIMEOUT`, inject nothing, and degrade to `url_context` with no visible failure — the quiet degradation this PR's fail-fast guardrails exist to prevent.

Given `timeoutSeconds: 900` on the revision, setting `URL_RESOLVER_TIMEOUT` explicitly here would make the budget a deploy-visible decision rather than an implicit library default.

---

MrOrz (commented):
Review of the sidecar config (the #134-merged shape). Two comments inline:

- `container-dependencies` makes a sidecar startup failure fatal to the whole revision, which contradicts the degrade-to-`url_context` design in #118 — and the image tag is mutable `:latest`.
- `URL_RESOLVER_TIMEOUT` is left at the client's 30s default, which is a whole-stream deadline for up to 20 URLs against a remote browser.

Checked and cleared: `CLOUDFLARE_KEEP_ALIVE_MS` rendering as `""` is fine — upstream parses it as `|| '60000'`, so the documented fallback holds. Resource sums and the rendered `container-dependencies` references also check out.

The substantive findings for this stack are on #118.

---

service.template.yaml
MrOrz: Fixed in f7829a8 — but dropping the dependency alone would not have fixed it, so the change goes further than suggested.

The dependency is not the operative gate; the sidecar's `startupProbe` is. Cloud Run's [runtime contract](https://docs.cloud.google.com/run/docs/container-contract) requires that "all containers within the instance need to be healthy" for an instance to serve, so a startup probe on `url-resolver` gates the whole revision's readiness regardless of what depends on it. Removing only the `container-dependencies` entry would have changed startup ordering and left the availability coupling exactly where it was.

What the dependency *did* do was make the probe mandatory — Cloud Run rejects the deployment with "Dependent container must have startup probe specified" — so both had to go together. The annotation now lists only `cloudsql-proxy`, and the sidecar has no probe.

Agreed that nothing is lost with it. As you note, it could only ever be `tcpSocket`, and the gRPC server binds before any browser is contacted, so passing it proved the port was open and never that a page could be rendered.

One limit worth stating rather than leaving implied: this does not make the sidecar wholly unable to affect the instance. A container that *exits* still terminates it, and that is not configurable away. What it removes is the failure-to-bind case — the one the mutable `:latest` tag makes reachable with no deploy on this side, which is the scenario your comment is really about.

Rendered the template through the workflow's `envsubst` allowlist and asserted on the parsed result: every `container-dependencies` reference resolves to a real container, `url-resolver` has no probe, and every container that is still depended on has one (or the deploy would be rejected). The ADR records the reasoning under "A failing sidecar must degrade, not take the service down".

---

service.template.yaml
MrOrz: Fixed in f7829a8 — `URL_RESOLVER_TIMEOUT` is now set explicitly next to the address, at **120s**.

Confirmed the mechanism before changing it: `resolve_urls()` passes the value straight to the streaming RPC (`stub.ResolveUrl(..., timeout=timeout)`), so it is a deadline for the whole stream, and the final list comprehension marks every URL absent from `results` as `TIMEOUT` — which the `ResolveStatus` docstring classes as "no signal". So the silent-degradation path is exactly as you describe.

The arithmetic does not clear at 30s: `DEFAULT_MAX_URLS = 20` at `SCRAPE_MAX_CONCURRENCY` 3 is ~7 waves, and even at the ADR's own ~5s per resolution that is ~35s — already past the deadline before accounting for the remote browser being slower than the in-process chromium the default was chosen against.

120s is ~7 waves at ~10s each plus headroom. The reasoning is in the template comment rather than just the commit, since the number is only defensible against those two constants. Worth revisiting if either changes.

Not raised, but it bounds the choice: this blocks the verifier's model call, so the timeout is also a cap on user-visible latency, not only a correctness knob. That is why it is 120s and not something much closer to the revision's 900s.

---

